PR #96, opened 2016-11-02
ignorePlurals: accept a list of ISO codes

LeoErcolanelli:
@ElPicador Is this what you were expecting? I used a class with inner classes to avoid creating multiple `.java` files, but that may be clearer than this.

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/objects/IgnorePlurals.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/objects/IgnorePlurals.java
@@ -1,0 +1,103 @@
+package com.algolia.search.objects;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.core.JsonGenerator;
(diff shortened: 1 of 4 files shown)


algoliasearch-common/src/main/java/com/algolia/search/objects/Get.java
```diff
@@ -0,0 +1,6 @@
+package com.algolia.search.objects;
+
+public class Get {
```
ElPicador: I would rename it `IgnorePlurals`

algoliasearch-common/src/main/java/com/algolia/search/objects/Get.java
```diff
@@ -0,0 +1,6 @@
...
+  public static class Boolean {}
```
ElPicador: I would more something like 

```
interface IgnorePlurals {}

class IgnorePluralsList extends IgnorePlurals {}
class IgnorePluralsBoolean extends IgnorePlurals {} 
```

But with better names

ElPicador (approved):

PR #100, opened 2016-11-08
Fix: distinct & removeStopWords

### LINKED ISSUE #99: Handling multiple response types for some settings' fields
The `distinct` field is defined in the `IndexSettings` class as `Integer`. However, it may also be returned as a `Boolean`. It raises a `com.fasterxml.jackson.databind.JsonMappingException` when Jackson tries to deserialize a `Boolean` as an `Integer`.

The same issue may occur with the `removeStopWords` setting that is defined as a `Boolean` but that may be returned from the REST API as an string array.

ElPicador:
Closes #99

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,8 @@ Changelog
 2.5.0 (2016-??-??)
 ------------------
 - Add searchFacet
+- Enable RemoveStopWords for boolean & List<String>
+- Enable Distinct for boolean and integer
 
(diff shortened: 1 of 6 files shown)


LeoErcolanelli:
I find this cool \o/ There is no drawbacks compared to using Object, but the equality is better handled here!

LeoErcolanelli:
I just fixed something you forgot in `Query.java`

PR #103, opened 2016-12-06
feat(sync): Add new retry strategy

ElPicador:


algoliasearch-common/src/main/java/com/algolia/search/APIClientConfiguration.java
```diff
@@ -15,6 +15,7 @@
   protected Map<String, String> headers;
   protected int connectTimeout;
   protected int readTimeout;
+  protected int hostDownTimeout;
```
PLNech, line 18: Don't we want a default value here? (And if yes, let's discuss what is the appropriate timeout!)

PLNech:
Apart from my comment on `hostDownTimeout` initialization, LGTM 👍

algoliasearch-common/src/main/java/com/algolia/search/APIClientConfiguration.java
ElPicador: Done!

PR #335, opened 2017-06-21
refactor(client): Split Index & AsyncIndex in multiple classes

ElPicador:
This PR only improves readability for the maintainers. 
It's a step before implementing the query rules

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
@@ -291,7 +291,7 @@ public class APIClient {
           TaskStatus.class)
       );
 
-      if (Objects.equals("published", status.getStatus())) {
+      if (java.util.Objects.equals("published", status.getStatus())) {
         return;
       }
       try {
@@ -653,7 +653,7 @@ public class APIClient {
 
   TaskSingleIndex batch(String indexName, List<BatchOperation> operations) throws AlgoliaException {
     //Special case for single index batches, indexName of operations should be null
-    boolean onSameIndex = operations.stream().allMatch(o -> Objects.equals(null, o.getIndexName()));
+    boolean onSameIndex = operations.stream().allMatch(o -> java.util.Objects.equals(null, o.getIndexName()));
     if (!onSameIndex) {
       throw new AlgoliaException("All operations are not on the same index");
     }
--- a/algoliasearch-common/src/main/java/com/algolia/search/AbstractIndex.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/AbstractIndex.java
@@ -1,13 +1,13 @@
 package com.algolia.search;
 
-abstract public class AbstractIndex<T> {
+public interface AbstractIndex<T> {
 
-  abstract public String getName();
+  String getName();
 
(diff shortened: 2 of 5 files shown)


clement-leprovost (commented):
Not sure to understand everything here: as far as I can tell, you refactored `AbstractIndex` from an abstract base class into an interface, `AsyncIndex` and `Index` from a concrete class into interfaces as well (consequently switching concrete methods to default methods), and introducing a new level of concrete classes. What's the purpose of this? Also, I can't see how this relates to upcoming features.

ElPicador:
@clement-leprovost I think the classes `Index` and `AsyncIndex` were too big & long. So I wanted to split them in multiple modules.
The only way to do this in Java 8 is to use interfaces, default implementations & abstract classes.

Does it make sense?

clement-leprovost (approved):
@ElPicador Yes. 🙂

ElPicador:
👍

mrmcc3:
Hello. It seems this change breaks basic interop/use from Clojure. I'm not 100% sure the technical reason why this causes an issue (could be something limiting with how Clojure interop works) but moving back to 2.10.0 resolves it.  

I realize ease of use from Clojure probably isn't a big factor but I thought i'd bring it up anyway.

ElPicador:
@mrmcc3 😢 . Sorry to have broke Clojure compatibility.

I'm not familiar with clojure, so could you provide me a way to reproduce your error so I can fix it?

ElPicador:
And feel free to open an issue on this

mrmcc3:
@ElPicador Sorry I lost track of this comment. I'm still on 2.10.0 but I thought i'd try 2.15.5 to see if the issue still exists and unfortunately it still gives me an error. 
```
CompilerException java.lang.IllegalArgumentException: Can't call public method of non-public class: public default com.algolia.search.objects.tasks.sync.TaskIndexing com.algolia.search.Objects.addObject(java.lang.Object) throws com.algolia.search.exceptions.AlgoliaException
```
I did a bit of googling and found similar errors from other java libraries. 
- https://stackoverflow.com/a/38060382
- https://github.com/netty/netty/issues/1780#issue-18493702

I'll admit this is all a bit over my head. But the comment from stuart in the second link seems to describe the problem pretty well.

ElPicador:
@mrmcc3 thanks for the link. We'll fix it quickly.

ElPicador:
Hi @mrmcc3,

We didn't manage to reproduce the error with the example code in the netty bug in java.
This code should reproduce but it passes:
```java
Class klass = Class.forName("com.algolia.search.Index");
APIClient client = new APIClient(null, null);
Index index = client.initIndex("test");

Method method;
Method methods[] = klass.getMethods();

for (Method method1 : methods) {
  method = method1;
  if ("addObject".equalsIgnoreCase(method.getName()) && method.getParameterCount() == 1) {
    method.invoke(index, "Test");
  }
}
```

We are far from clojure developers. Would you mind giving us a clojure code that fails, so we can add a test and try to fix it.

Thanks!

mrmcc3:
Yeah weird that should reproduce it. I opened an issue with some clojure code that fails for me

ElPicador:
Thanks, let's continue this discussion on the issue then!

PR #336, opened 2017-06-22
feat(query_rules): Add query rules

ElPicador:


clement-leprovost:
My fault, as I forgot to state that explicitly: The naming should be coordinated with the other API clients, and we went for "rules" alone, not "query rules" for method naming. So this would be `saveRule` instead of `saveQueryRules`. Sorry for that... 😬

algoliasearch-common/src/main/java/com/algolia/search/inputs/query_rules/QueryRule.java
```diff
@@ -0,0 +1,166 @@
+package com.algolia.search.inputs.query_rules;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+public class QueryRule {
```
clement-leprovost: I wonder if this class, which closely mirrors the JSON schema, provides the best developer experience. I chose `if` and `then` in the schema because it makes rules data relatively straightforward to read. But when it comes to code... maybe it's not the best idea I ever had. 😆 => I would go for `Condition` and `Consequence`.

IMHO, it would also be OK to use public fields for the sub-objects, especially if they are declared `final`, as they are mandatory anyway.

So, instead of `rule.getIfQueryRule().setPattern("foo bar")` you could write `rule.condition.setPattern("foo bar")`.

This doesn't apply to the description as it's optional and atomic (i.e. not an aggregate).

algoliasearch-common/src/main/java/com/algolia/search/responses/SearchQueryRuleResult.java
```diff
@@ -0,0 +1,49 @@
+package com.algolia.search.responses;
+
+import com.algolia.search.inputs.query_rules.QueryRule;
+
+import java.util.List;
+
+public class SearchQueryRuleResult {
```
clement-leprovost: I know we use the same pattern for the regular search, but I am concerned about forward-compatibility here. Let's say we add another field in the response that is not supported by this class: what will happen? Parse error? (For example, we could add debug fields if certain options are activated.)

clement-leprovost (commented):
There are specific challenges due to the strongly typed nature of Java that I did not anticipate. => Let's discuss them.

ElPicador:
@clement-leprovost done here also

algoliasearch-common/src/main/java/com/algolia/search/objects/SynonymQuery.java
```diff
@@ -25,12 +29,21 @@ public SynonymQuery setQuery(String query) {
     return this;
   }
 
-  public List<String> getTypes() {
-    return types;
+  public String getAnchoring() {
```
clement-leprovost: Why is there anchoring on `SynonymQuery`? That notion exists in the rule, but not in the synonyms AFAIK.

algoliasearch-common/src/main/java/com/algolia/search/objects/SynonymQuery.java
```diff
@@ -25,12 +29,21 @@ public SynonymQuery setQuery(String query) {
...
+    return this;
+  }
+
+  public String getContext() {
+    return context;
   }
 
-  public SynonymQuery setTypes(List<String> types) {
-    this.types = types;
+  public SynonymQuery setContext(String context) {
```
clement-leprovost: Same as above: synonyms don't have a context (rules do).

clement-leprovost (changes requested):
I may be missing something, but there seems to have been unwanted copy-pasting between rules and synonyms.

algoliasearch-common/src/main/java/com/algolia/search/objects/SynonymQuery.java
ElPicador: Good catch!

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -810,13 +811,13 @@ TaskSingleIndex partialUpdateObjects(String indexName, List<Object> objects) thr
...
     AlgoliaRequest<BrowseResult> algoliaRequest = new AlgoliaRequest<>(
-      HttpMethod.GET,
+      HttpMethod.POST,
       true,
       Arrays.asList("1", "indexes", indexName, "browse"),
       BrowseResult.class,
       klass
     )
-      .setParameters(query.setCursor(cursor).toQueryParam());
+      .setData(query.setCursor(cursor));
```
clement-leprovost, line 991: What's it got to do with Query Rules?

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -830,11 +831,88 @@ SearchFacetResult searchForFacetValues(String indexName, String facetName, Strin
       Arrays.asList("1", "indexes", indexName, "facets", facetName, "query"),
       SearchFacetResult.class
     )
-      .setData(new Search(query));
+      .setData(query);
```
clement-leprovost: Same here.

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -830,11 +831,88 @@ SearchFacetResult searchForFacetValues(String indexName, String facetName, Strin
...
+
+  Task batchRules(String indexName, List<Rule> queryRules, Boolean forwardToReplicas, Boolean replaceExistingQueryRules) throws AlgoliaException {
+    Task task = httpClient.requestWithRetry(
+      new AlgoliaRequest<>(
+        HttpMethod.POST,
+        false,
+        Arrays.asList("1", "indexes", indexName, "rules", "batch"),
+        Task.class
+      )
+        .setParameters(ImmutableMap.of("forwardToReplicas", forwardToReplicas.toString(), "replaceExistingQueryRules", replaceExistingQueryRules.toString()))
```
clement-leprovost: It's `clearExistingRules`, not `replaceExistingQueryRules`.

algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
```diff
@@ -300,7 +302,7 @@ public String generateSecuredApiKey(@Nonnull String privateApiKey, @Nonnull Quer
       try {
         result = status.get();
       } catch (CancellationException | InterruptedException | ExecutionException e) {
-        //If the future was cancelled or the thread was interrupted or future completed exceptionally
+        //Condition the future was cancelled or the thread was interrupted or future completed exceptionally
```
clement-leprovost: Dubious. Probably search & replace side effect. :)

algoliasearch/src/test/java/com/algolia/search/ApacheHttpClientTest.java
```diff
@@ -116,7 +115,7 @@ public void run() {
   @Test
   public void shouldHandleSNI() throws Exception {
     APIClient client = build(APPLICATION_ID + "-1.algolianet.com");
-    assertThat(client.listKeys()).isNotEmpty();
+    assertThat(client.listApiKeys()).isNotEmpty();
```
clement-leprovost: Unrelated to Query Rules?!

clement-leprovost (changes requested):
Sorry, but there are still some fixes required. :/

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
ElPicador: It simplifies the handling of the object `Query` that is needed for the query rules

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
ElPicador: same

algoliasearch/src/test/java/com/algolia/search/ApacheHttpClientTest.java
ElPicador: yes, it's only a fix because `listKeys` is deprecated

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
ElPicador: you are right thanks!

algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
ElPicador: thanks

clement-leprovost (approved):
LGTM 👍

PR #338, opened 2017-06-30
Do not assume versions.properties is the only one on the classpath

adamretter:
As this is an API that is meant to be used in other applications, it should not assume that the `version.properties` file (needed for the UserAgent in the requests), will be the only such one on the classpath.

My application was reporting the wrong API client version to Algolia, because another 3rd-party dependency that I had included (isorelax) was making the same mistake as the Algolia client, and also had placed its own `version.properties` on the classpath. Algolia was picking up the `version.properties` file of isorelax instead of its own.

To solve this, in this PR I have moved the Algolia `version.properties` into the `com.algolia.search` so that it does not clash with anything else.


ElPicador:
Nice catch. And thanks for adding a test.

PR #339, opened 2017-07-03
feat(query): Add request options to search/read methods

ElPicador:


algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -54,6 +55,20 @@
   }
 
   /**
+   * List all existing indexes
+   *
+   * @return A List of the indices and their metadata
+   * @throws AlgoliaException
+   */
+  public List<Index.Attributes> listIndices(RequestOptions requestOptions) throws AlgoliaException {
```
LeoErcolanelli: I may be missing something, but requestOptions is not used, right?

ElPicador: You are right, this PR was only the start of a POC.
So I didn't implement everything yet

LeoErcolanelli: My bad, I missed the "early" before "implementation" in Clément's mail

clement-leprovost (approved):
LGTM 👍 

It's interesting that you chose to go for a `@NonNull` argument, and specify `RequestOptions.empty`. Is there a specific reason? It's true that explicitly passing "empty" makes the code a tad more readable. :)

ElPicador:
The second thing I hate the most are `null` and `NullPointerException` so I try to enforce no `null` and use different patterns (empty objects, `Option`, etc.).

PR #342, opened 2017-07-07
fix(browse): Fix browse

ElPicador:


clement-leprovost (approved):
LGTM.

Maybe we could remove `isFirstRequest` and only keep a condition on `currentIterator == null`.

ElPicador:
You're right. I've removed it.

PR #358, opened 2017-08-22
feat(delete): Add delete by query

ElPicador:


algoliasearch-common/src/main/java/com/algolia/search/Index.java
```diff
@@ -1433,44 +1433,74 @@ default Task batchRules(@Nonnull List<Rule> rules, @Nonnull RequestOptions reque
...
+   * @param requestOptions Options to pass to this request
+   * @throws AlgoliaException
+   */
+  default Task deleteBy(@Nonnull Query query, @Nonnull RequestOptions requestOptions) throws AlgoliaException {
+    return getApiClient().deleteBy(getName(), query, requestOptions);
+  }
+
+  /**
    * Delete records matching a query, internally uses browse
+   * Deprecated use deleteBy
```
clement-leprovost: Doesn't Javadoc have a `@deprecated` tag? I know we already have an annotation, but if we put something in the comment, why not use it?

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -968,6 +968,20 @@ void deleteByQuery(String indexName, Query query, int batchSize, RequestOptions
...
+  Task deleteBy(String indexName, Query query, RequestOptions requestOptions) throws AlgoliaException {
+    query = query == null ? new Query() : query;
+    AlgoliaRequest<Task> algoliaRequest = new AlgoliaRequest<>(
+      HttpMethod.POST,
+      false,
+      Arrays.asList("1", "indexes", indexName, "deleteByQuery"),
+      requestOptions,
+      Task.class
+    )
+      .setData(query);
```
clement-leprovost, line 980: What will `setData` do? As per the [REST API Reference](https://www.algolia.com/doc/rest-api/search/#delete-by-query), the API expects a JSON object with a `params` attribute containing all parameters as a URL-encoded string.

algoliasearch-common/src/test/java/com/algolia/search/integration/sync/SyncDeleteByTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @SuppressWarnings("deprecation")
+  @Test
+  public void deleteByQuery() throws AlgoliaException {
+    Index<AlgoliaObject> index = client.initIndex("index1", AlgoliaObject.class);
+
+    List<AlgoliaObject> objects = IntStream.rangeClosed(1, 10).mapToObj(i -> new AlgoliaObject("name" + i, i)).collect(Collectors.toList());
+    index.addObjects(objects).waitForCompletion();
+
+    index.deleteByQuery(new Query(""));
```
clement-leprovost, line 45: Shouldn't you wait for completion here? Delete by query is asynchronous as all write operations; you must wait on the task ID.

ElPicador: It's the old deleteByQuery, so no tasks here

algoliasearch-common/src/main/java/com/algolia/search/Index.java
ElPicador: Yep, adding it.

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
ElPicador: `search` also works with a plain JSON objects, does it works here also?

algoliasearch-common/src/test/java/com/algolia/search/integration/sync/SyncDeleteByTest.java
ElPicador: you are right

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
clement-leprovost: Looking at the code, it seems so, but it's not documented... :/

clement-leprovost (approved):
LGTM 👍

PR #397, opened 2017-11-16
fix(output): Replace Integers by Longs

ElPicador:
cc: @raphi


GregoryOrciuch:
@ElPicador 

Hi, 
For future, please consider use of  "@Deprecated" for changing types of crucial methods in algolia api like "getEntries()", then staging such changes in time. We had to spent the time to investigate why our product was not compiling anymore.  Thanks for future consideration and all your hard work :)

PR #412, opened 2017-12-08
fix(exposePartialUpdateObjectNoCreate): Expose the PartialUpdateObjec…

robertmogos:
…tNoCreate to the client

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
@@ -1037,7 +1037,7 @@ public class AsyncAPIClient {
 
   CompletableFuture<AsyncTaskSingleIndex> partialUpdateObjects(
       String indexName, List<Object> objects, RequestOptions requestOptions) {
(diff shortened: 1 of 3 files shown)


seafoox:
Great thanks @robertmogos

spinach (approved):

algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
```diff
@@ -1036,14 +1037,19 @@ public String generateSecuredApiKey(
 
   CompletableFuture<AsyncTaskSingleIndex> partialUpdateObjects(
       String indexName, List<Object> objects, RequestOptions requestOptions) {
+    return partialUpdateObjects(indexName, objects, requestOptions, false);
```
PLNech: Shouldn't it be `true` by default? Else we introduce a breaking change in the default behavior of `partialUpdateObjects` (used to create by default, now would _not_ create by default) 🤔

PLNech (changes requested):

PLNech (approved):

PR #442, opened 2018-01-15
fix(client): Rename `client.updateKey` => `client.updateApiKey`

### LINKED ISSUE #423: rename client.updateKey client.updateApiKey

ElPicador:
Closes #423

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
@@ -342,7 +342,7 @@ public class APIClient {
             .setData(key));
   }
 
-  /** Deprecated: use updateApiKey */
+  /** Deprecated: use {@link #updateApiKey(String, ApiKey)} */
   @Deprecated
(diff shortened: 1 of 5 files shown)


algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -342,7 +342,7 @@ public CreateUpdateKey addApiKey(@Nonnull ApiKey key, @Nonnull RequestOptions re
             .setData(key));
   }
 
-  /** Deprecated: use updateApiKey */
+  /** Deprecated: use {@link #updateApiKey(String, ApiKey)} */
```
aseure, line 345: Javadoc seems cool. Is it part of Javadoc to also include the `Deprecated` note (as it's already an annotation below)?

aseure (approved):

julienbourdeau (approved):

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
ElPicador: No it's only me.

PR #456, opened 2018-02-13
fix(client): Add one deleteIndex method (#427)

### LINKED ISSUE #427: Implement client.deleteIndex(indexName, requestOptions)

aseure:
Closes #427 

Apparently, the method with `RequestOptions` wasn't missing. So here, I'm just adding the one without it, to stay consistent with other methods.

# Question
However, I have a question: I've also added the `@Nonnull` annotations on the existing `deleteIndex` method. Will it break existing implementations? It will prevent people passing `null` from compiling but this is really a useful check, it's easy to fix and it shouldn't happen in the first place. WDYT?

# Test plan
As the first method is already tested for both sync and async clients, I have added no tests.

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
@@ -233,6 +233,37 @@ public class APIClient {
   }
 
   /**
+   * Delete an existing index
+   *
+   * @param indexName The index name that will be deleted
+   * @return The associated task
+   */
+  public Task deleteIndex(@Nonnull String indexName) throws AlgoliaException {
+    return deleteIndex(indexName, RequestOptions.empty);
+  }
+
+  /**
+   * Delete an existing index
+   *
+   * @param indexName The index name that will be deleted
+   * @param requestOptions Options to pass to this request
+   * @return The associated task
+   */
+  public Task deleteIndex(@Nonnull String indexName, @Nonnull RequestOptions requestOptions) throws AlgoliaException {
+    Task result =
+            httpClient.requestWithRetry(
+                    new AlgoliaRequest<>(
+                            HttpMethod.DELETE,
+                            false,
+                            Arrays.asList("1", "indexes", indexName),
+                            requestOptions,
+                            Task.class));
+
+    return result.setAPIClient(this).setIndex(indexName);
+  }
+
+
+  /**
    * Return 10 last log entries.
    *
    * @return A List<Log>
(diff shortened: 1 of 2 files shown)


algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -232,6 +232,24 @@ public Task copyIndex(
     return copyIndex(srcIndexName, dstIndexName, scopes, RequestOptions.empty);
   }
 
+  public Task deleteIndex(@Nonnull String indexName) throws AlgoliaException {
```
ElPicador: Could you add the javadoc for those 2 methods?

ElPicador (changes requested):
The method should also be added in AsyncClient.

It's ok to add the `@Nonnull` it was only an internal method that is public now.

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
aseure: Yes right, totally forgot. Will do.

aseure:
@ElPicador I think we're good now. Let me know.

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -232,6 +232,39 @@ public Task copyIndex(
     return copyIndex(srcIndexName, dstIndexName, scopes, RequestOptions.empty);
   }
 
+  /**
+   * Delete an existing index
+   *
+   * @param indexName The index name that will be deleted
+   * @return The associated task
+   * @throws AlgoliaException
```
ElPicador: remove if you do not put a comment

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -232,6 +232,39 @@ public Task copyIndex(
...
+    return deleteIndex(indexName, RequestOptions.empty);
+  }
+
+  /**
+   * Delete an existing index
+   *
+   * @param indexName The index name that will be deleted
+   * @param requestOptions Options to pass to this request
+   * @return The associated task
+   * @throws AlgoliaException
```
ElPicador: same

algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
```diff
@@ -648,7 +648,11 @@ public String generateSecuredApiKey(
   }
 
   /** Package protected method for the Index class */
-  CompletableFuture<AsyncTask> deleteIndex(String indexName, RequestOptions requestOptions) {
+  CompletableFuture<AsyncTask> deleteIndex(@Nonnull String indexName) {
```
ElPicador: the method should be public not protected anymore, like above

ElPicador (changes requested):

algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
aseure: Ok I wasn't sure 👍

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
aseure: Ok I'll remove it.

ElPicador (approved):

PR #466, opened 2018-03-04
Adds the 'facet_stats' attribute to the SearchResult object

wietsevenema:
Hi,

The search results from the api contain a map with facet stats per (numerical) facet, but I can't access them. I would like to access them. 

This PR adds the attribute to the SearchResult object. I thought it was best to keep it a map with string keys since I don't want it to break with api updates (new stats keys). 

I ran the tests locally and got `APPLICATION_ID is not defined or empty` errors. I am assuming that is for an integration test I don't have? 

If you have any requests, please let me know.

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/responses/SearchResult.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/responses/SearchResult.java
@@ -20,6 +20,8 @@ public class SearchResult<T> implements Serializable {
 
   private Map<String, Map<String, Long>> facets;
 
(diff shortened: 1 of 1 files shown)


ElPicador:
Thanks for the PR.
The facet_stats is a Map<String, Float>. As I can not edit your PR, I've made another one: https://github.com/algolia/algoliasearch-client-java-2/pull/467.

So I'm closing this one.

wietsevenema:
thanks 👍

ElPicador:
It's available in the `2.16.3` release

PR #476, opened 2018-03-30
fix(query): Fix AroundRadius serialization

ElPicador:


algoliasearch-common/src/main/java/com/algolia/search/objects/Query.java
```diff
@@ -289,15 +297,27 @@ public String toParam() {
...
     return this;
   }
 
-  @JsonIgnore
+  /**
+   * @deprecated use setAroundRadius(AroundRadius)
+   * @return
+   */
+  @Deprecated()
+  public Query setAroundRadius(Integer aroundRadius) {
```
JonathanMontane: Apparently, the tests are failing because of a conflict in the setters for `AroundRadius`.
One of these two methods should be annotated with `@JsonIgnore`.

PR #479, opened 2018-04-09
fix: Prevent saving a rule with an empty objectID

aseure:
Follow up of https://github.com/algolia/algoliasearch-client-go/issues/397.


ElPicador (approved):
I'll add a test on the nullity of `queryRuleID`. It's Java so everything can be null.

aseure:
@ElPicador Not sure we want to add a test for every parameter x every method that can be `null`. Plus, we already have the `@Nonnull` annotations anyway.

PR #482, opened 2018-04-30
feat: Deprecate API keys methods on Index in favor of Client ones

aseure:


ElPicador (approved):

aseure:
@ElPicador Could you help us understand why the AppEngine test in failing? We're a bit lost here. Thank you in advance.

PR #484, opened 2018-05-16
Update common files

julienbourdeau:
:gift:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,21 @@
+| Q                 | A
+| ----------------- | ----------
+| Bug fix?          | yes/no
+| New feature?      | yes/no    <!-- please update the /CHANGELOG.md file -->
+| BC breaks?        | no     
+| Related Issue     | Fix #...  <!-- will close issue automatically, if any -->
+| Need Doc update   | yes/no
+
+
+## What was changed
+
+<!-- 
+    Please give the reviewers and contributors 
+    as many details as possible to understand your changes
+-->
+
+## Why it was changed
+
+<!-- 
+    Please explain why this change is necessary or interesting 
+-->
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+
+
+## What was changed
+
+<!-- 
+    Please give the reviewers and contributors 
+    as many details as possible to understand your changes
+-->
+
+## Why it was changed
```
aseure, line 17: `s/Why it was changed/Why was it changed?`

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+| Q                 | A
+| ----------------- | ----------
+| Bug fix?          | yes/no
+| New feature?      | yes/no    <!-- please update the /CHANGELOG.md file -->
+| BC breaks?        | no     
+| Related Issue     | Fix #...  <!-- will close issue automatically, if any -->
+| Need Doc update   | yes/no
+
+
+## What was changed
```
aseure, line 10: We should either use `What has changed` or `What was changed?` (with the question mark) if I'm not wrong. cc @codeharmonics

codeharmonics:
Hey @aseure, I agree: "What was changed" sounds awkward. It's passive. But "What has changed" is even more passive, as if something has changed but nobody is responsible. 

There is "What did you change?" or "What changes did you make?" or "Describe your change" or "Please describe your change (add as much detail as necessary to understand your code)".

If you don't want to say "you", to avoid it sounding like an interrogation, then "What is the change?" or "Describe the change".

For the next one, I'd go with:

"What problem is this fixing?"
with "(Please include everything needed to understand the problem, its context and consequences, and, if possible, how to recreate it)"

My 2 cents.

PR #493, opened 2018-07-03
Ab testing

aseure:


algoliasearch-common/src/main/java/com/algolia/search/Defaults.java
```diff
@@ -19,7 +22,9 @@
   ObjectMapper DEFAULT_OBJECT_MAPPER =
       new ObjectMapper()
           .enable(OBJECT_MAPPER_DEFAULT_FEATURE)
-          .disable(OBJECT_MAPPER_DEFAULT_DESERIALIZATION_FEATURE);
+          .disable(OBJECT_MAPPER_DEFAULT_DESERIALIZATION_FEATURE)
+          .registerModule(new JavaTimeModule())
```
ElPicador: careful with this.
Users can override the default object mapper. So the `JavaTimeModule` might not present.

algoliasearch-common/src/main/java/com/algolia/search/Defaults.java
```diff
@@ -19,7 +22,9 @@
...
+          .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
```
ElPicador: Same comment

algoliasearch-common/src/main/java/com/algolia/search/http/AlgoliaRequestKind.java
```diff
@@ -0,0 +1,7 @@
+package com.algolia.search.http;
+
+public enum AlgoliaRequestKind {
+  SEARCH,
```
ElPicador: I would prefix the values by the product.
So here:
`SEARCH_API_SEARCH`
`SEARCH_API_WRITE`
`ANALYTICS_READ`

algoliasearch-common/src/main/java/com/algolia/search/inputs/analytics/ABTest.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+
+public class ABTest {
+  // Properties available at construction time
+  public String name;
+  public List<Variant> variants;
+  public ZonedDateTime endAt;
+
+  // Properties set by the AB Testing API
+  @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
```
ElPicador: why all the `@JsonProperty(access = JsonProperty.Access.WRITE_ONLY)` ?

algoliasearch-common/src/main/java/com/algolia/search/inputs/analytics/ABTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+public class ABTest {
+  // Properties available at construction time
+  public String name;
```
ElPicador: All the POJOs are beans, so for consistency, put all fields private and generate getters/setters if needeed.
For example: https://github.com/algolia/algoliasearch-client-java-2/blob/master/algoliasearch-common/src/main/java/com/algolia/search/responses/FacetHit.java

algoliasearch-common/src/main/java/com/algolia/search/inputs/analytics/ABTest.java
```diff
@@ -0,0 +1,38 @@
+package com.algolia.search.inputs.analytics;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+public class ABTest {
```
ElPicador: don't forget to add `Serializable` on all POJOs

algoliasearch-tests/src/test/java/com/algolia/search/integration/common/async/AsyncABTestingTest.java
```diff
@@ -0,0 +1,96 @@
...
+    inserted = analytics.GetABTest(inserted.abTestID).get();
+    assertThat(inserted.status).isEqualTo("stopped");
+
+    waitForCompletion(analytics.DeleteABTest(inserted.abTestID));
+    abTests = analytics.GetABTests(0, 10).get();
+    assertThat(abTests.count).isEqualTo(0);
+  }
+
+  public static void compareABTests(ABTest abtest, ABTest inserted) {
+    for (int i = 0; i < 2; i++) {
```
ElPicador: maybe do a check that both have variants of size 2

ElPicador (commented):
Don't forget to run `mvn com.coveo:fmt-maven-plugin:format` before commiting. It'll format the code for you

algoliasearch-common/src/main/java/com/algolia/search/Defaults.java
aseure: I did not know that. What would be the best way to enforce its use then. I've checked extensively and the `JavaTimeModule` is the correct way to format Java8 time related classes. I'd like to stick with this rather that workarounds.

algoliasearch-common/src/main/java/com/algolia/search/http/AlgoliaRequestKind.java
aseure: Good idea, changing it.

algoliasearch-common/src/main/java/com/algolia/search/inputs/analytics/ABTest.java
aseure: This is one way to make Jackson **only** deserialize a property (i.e. force it not to serialize those fields). In our case, it is useful because most of the fields are read-only and produced by the Analytics API upon read. When adding/editing tests, the user can only write the top fields that are the only ones accessible as RW.

I'll remove it anyway following your recommendation to use getter/setter rather than public attributes. Because if I only provide the setter, it has the same effect as this annotation.

For the record, I've found all those informations here:
http://www.baeldung.com/jackson-field-serializable-deserializable-or-not.

algoliasearch-common/src/main/java/com/algolia/search/inputs/analytics/ABTest.java
aseure: Yes, I get the consistency reason and after reading again few things, it feels really wrong to use public fields. Thanks for the review about that. I'll remove all the public fields and just add getters (apparently, you don't need both getter and setter with Jackson, because getter=serialization+deserialization whereas setter=deserialization only, still according to http://www.baeldung.com/jackson-field-serializable-deserializable-or-not).

For the record, I think it's the opposite (all JavaBeans are POJOs but not all POJOs are JavaBeans) according to what I found on internet around this subject. My brain right now:
giphy

algoliasearch-common/src/main/java/com/algolia/search/inputs/analytics/ABTest.java
aseure: Oh right, good catch. Will do 👍

algoliasearch-tests/src/test/java/com/algolia/search/integration/common/async/AsyncABTestingTest.java
aseure: My bad, I've extracted the comparison code from the test function and forgot to include the assertion that is still in there. I'll move it back. Good catch, thank you.

aseure:
I have addressed all your comments @ElPicador, besides the one regarding the defaults for the `ObjectMapper` where I'll wait for your input. Let me know if I missed something.

aseure:
@ElPicador I've addressed the very last comment: I've removed the dependency on the ObjectMapper new Java8 Time serializers and made a custom one instead. Ready to merge on my side.

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -1362,6 +1380,61 @@ Task batchRules(
     return task.setAPIClient(this).setIndex(indexName);
   }
 
+  public TaskABTest AddABTest(ABTest abtest) throws AlgoliaException {
```
ElPicador: methods in Java starts with lower case.
Seems all your methods are starting with capital case

aseure: Indeed, will change it.

aseure: Done. Waiting for Travis to run. Let me know if you see something else.

ElPicador (approved):

PR #497, opened 2018-08-01
Introduce MultiCluster Management

julienbourdeau:


julienbourdeau (commented):
Thank you for the `RequestOption.empty` fix 🙏 

It would be great to add more tests on AsyncMultiCluster (I only made a super simple one)

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -1435,6 +1437,123 @@ public ABTests getABTests(int offset, int limit) throws AlgoliaException {
...
+    return this.getUserID(userID, RequestOptions.empty);
+  }
+
+  public UserID getUserID(@Nonnull String userID, RequestOptions requestOptions)
+      throws AlgoliaException {
+    String encodedUserID = "";
+
+    try {
+      encodedUserID = URLEncoder.encode(userID, "UTF-8");
+    } catch (UnsupportedEncodingException ignore) {
```
Ant-hem: Shouldn't we throw an exception here?

algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
```diff
@@ -1357,4 +1359,116 @@ public String generateSecuredApiKey(
...
+    return this.getUserID(userID, RequestOptions.empty);
+  }
+
+  public CompletableFuture<UserID> getUserID(
+      @Nonnull String userID, RequestOptions requestOptions) {
+    String encodedUserID = "";
+
+    try {
+      encodedUserID = URLEncoder.encode(userID, "UTF-8");
+    } catch (UnsupportedEncodingException ignore) {
```
Ant-hem: Same comment as above.

algoliasearch-common/src/main/java/com/algolia/search/responses/DeleteUserID.java
```diff
@@ -0,0 +1,24 @@
...
+
+  @SuppressWarnings("unused")
+  public DeleteUserID setDeletedAt(String deletedAt) {
+    this.deletedAt = deletedAt;
+    return this;
+  }
+
+  @Override
+  public String toString() {
+    return "DeleteUserID{" + "deletedAt='" + deletedAt + '\'' + '}';
```
Ant-hem, line 22: It's a nit but maybe using String.Format would be more clearer ?

Ant-hem (changes requested):

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
aseure: I agree with you @Ant-hem, I'll make those throw an `AlgoliaException` wrapping the `UnsupportedEncodingException` rather than ignoring (bad enough to prevent the HTTP call). Let us know if it was on purpose @julienbourdeau.

algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
aseure: Same answer as above ^^

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
aseure: Good catch @Ant-hem

algoliasearch-common/src/main/java/com/algolia/search/AsyncAPIClient.java
aseure: And wrapping the error in a `CompletableFuture` as the `getUserID` of the async client cannot throw.

algoliasearch-common/src/main/java/com/algolia/search/responses/DeleteUserID.java
aseure: Agree but as it seems to be written like this everywhere else in the code, I think @julienbourdeau followed the same pattern for consistency which I'd agree with.

Ant-hem: Yes, you're right, we can leave it like this for consistency. :)

Ant-hem (approved):

PR #498, opened 2018-08-07
fix: Prevent double-close of response's body when DEBUG logging is enabled

aseure:


aseure:
@ElPicador Let me know what you think, especially in term of performances if you see any non-negligible impact. Thanks again for your help.

ElPicador (approved):

aseure:
Released in [2.18.1](https://github.com/algolia/algoliasearch-client-java-2/blob/master/ChangeLog.md#2181-2018-08-07)

julienbourdeau (commented):
Ideally, I wish we didn't have to use the return from `logResponse` and that this log function has no effect on the state, but I guess you did it this way for a reason ^^
👍 👍

aseure:
@julienbourdeau I thought the same. I've tried a few things to wrap the internal body response's content of the `HttpEntity` into something re-readable and re-closeable but it didn't work out. This change was our last idea with @ElPicador. It has the advantage of having no overhead if you don't enable the debug logging level.

PR #506, opened 2018-09-24
Fix facet filters

### LINKED ISSUE #505: Change facetFilters type from List<String> to List<Object>
The current implementation does not support OR operations because the type of `facetFilters` is `List<String>`, which makes it impossible to pass lists of `List<String>`.

To fix this, I'll probably need to change the type for `List<Object>`. @ElPicador Could you confirm that there's no better solution?

aseure:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | yes    
| Related Issue     | Fix #505 
| Need Doc update   | yes

## Describe your change

This change introduces a new `FacetFilters` class to replace the `List<String>` type of the `facetFilters` parameter used in the `Query` and in the `ConsequenceParams` (extending `Query`).

Currently, because a user can only define `facetFilters` as a list of strings, filters can only be `AND`ed. To be able to `OR` filters, the `facetFilters` should accept lists of list of strings as well. This is what this change is all about. Using the new `FacetFilters` abstract class that is introduced, and its `FacetFiltersAsListOfString`/`FacetFiltersAsListOfList` concrete types, one can easily set the `facetFilters` parameter of a `Query`/`ConsequenceParams` using `List<String>` or `List<List<String>>` types. As any other multi-type object of this Java client, deserialization of object returns an `Object` that can be cast to one of the concrete classes, however, this shouldn't be used that often as the `facetFilters` parameter is only sent but never returned by the API, but anyway, it's doable.

@ElPicador As it's one of the rare times I add custom (de)serializers myself, could you tell me if there's better/more idiomatic ways to write them?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ hs_err_pid*
 *.iml
 .DS_Store
 *.logQueryTest
+
+.factorypath
--- a/algoliasearch-common/src/main/java/com/algolia/search/objects/FacetFilters.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/objects/FacetFilters.java
@@ -1,0 +1,68 @@
+package com.algolia.search.objects;
+
+import com.algolia.search.Defaults;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.List;
+
+@JsonDeserialize(using = FacetFiltersJsonDeserializer.class)
+@JsonSerialize(using = FacetFiltersJsonSerializer.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public abstract class FacetFilters implements Serializable {
+
+  public static FacetFilters ofList(List<String> filters) {
+    return new FacetFiltersAsListOfString(filters);
+  }
+
+  public static FacetFilters ofListOfList(List<List<String>> filters) {
+    return new FacetFiltersAsListOfList(filters);
+  }
+
+  @JsonIgnore
+  public abstract Object getInsideValue();
+}
+
+class FacetFiltersJsonDeserializer extends JsonDeserializer {
+
+  @Override
+  public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+    List list = p.readValueAs(List.class);
+    if (list.isEmpty()) {
+      return new FacetFiltersAsListOfString();
+    }
+    if (list.get(0) instanceof String) {
+      return new FacetFiltersAsListOfString(list);
+    }
+    return new FacetFiltersAsListOfList(list);
+  }
+}
+
+class FacetFiltersJsonSerializer extends JsonSerializer<FacetFilters> {
+
+  @Override
+  public void serialize(FacetFilters value, JsonGenerator gen, SerializerProvider serializers)
+      throws IOException {
+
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -19,3 +19,5 @@ hs_err_pid*
 *.iml
 .DS_Store
 *.logQueryTest
+
+.factorypath
```
ElPicador, line 23: what is this?

algoliasearch-common/src/main/java/com/algolia/search/objects/FacetFilters.java
```diff
@@ -0,0 +1,76 @@
...
+  public Object deserialize(JsonParser p, DeserializationContext ctxt)
+      throws IOException, JsonProcessingException {
+    List list = p.readValueAs(List.class);
+    if (list.isEmpty()) {
+      return new FacetFiltersAsListOfString();
+    }
+    if (list.get(0) instanceof String) {
+      return new FacetFiltersAsListOfString(list);
+    }
+    return new FacetFiltersAsListOfList(list);
```
ElPicador, line 43: I would add a check here that it's indeed a list of list

algoliasearch-common/src/main/java/com/algolia/search/objects/FacetFilters.java
```diff
@@ -0,0 +1,76 @@
...
+      gen.writeStartArray();
+      for (List<String> l : filters.getFilters()) {
+        gen.writeStartArray();
+        for (String f : l) {
+          gen.writeString(f);
+        }
+        gen.writeEndArray();
+      }
+      gen.writeEndArray();
+    }
```
ElPicador, line 66: missing else or at least an exception thrown

ElPicador (commented):
I think you forgot the tests :P

.gitignore
aseure: Apparently, files generated by Maven upon IntelliJ importing of a new project. First time I see those as well.

algoliasearch-common/src/main/java/com/algolia/search/objects/FacetFilters.java
aseure: As the `FacetFiltersAsListOfList` is expecting a `List<List<String>>`, I supposed this would throw if `list` was not of type `List<List<String>>`, hence validating the expected behaviour. Does it look wrong to you?

Otherwise, do you want me to check that the first element is of type `List<String>` and otherwise throw a custom exception?

algoliasearch-common/src/main/java/com/algolia/search/objects/FacetFilters.java
aseure: Good catch, thank you 👍

aseure:
@ElPicador Forgot to *commit* the tests actually. Will do.

algoliasearch-common/src/main/java/com/algolia/search/objects/FacetFilters.java
ElPicador: I would throw a meaning full exception instead on relying on what the JVM will generate

ElPicador:
> @ElPicador Forgot to _commit_ the tests actually. Will do.

I see what you did there :suspect:

algoliasearch-common/src/main/java/com/algolia/search/objects/FacetFilters.java
```diff
@@ -0,0 +1,81 @@
...
+
+class FacetFiltersJsonSerializer extends JsonSerializer<FacetFilters> {
+
+  @Override
+  public void serialize(FacetFilters value, JsonGenerator gen, SerializerProvider serializers)
+      throws IOException {
+    if (value instanceof FacetFiltersAsListOfString) {
+      FacetFiltersAsListOfString filters = (FacetFiltersAsListOfString) value;
+      gen.writeStartArray();
+      for (String f : filters.getFilters()) {
```
Ant-hem: Maybe you could use the objectMapper.readerFor method instead of the writer to have a more simple code : 
```
            ObjectReader reader = objectMapper.readerFor(new TypeReference<FacetFiltersAsListOfString>() {
            });
            FacetFiltersAsListOfString ret = reader.readValue(node);
```

PR #507, opened 2018-09-25
Query rules v2

Ant-hem:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes 
| BC breaks?        | no     
| Related Issue     | no
| Need Doc update   | yes

## Describe your change

Added the Enabled feature for Query rules v2
Added the Demote (Hide) test case
Added the TimeFrame test case
Added the Query Edits test case and custom Deserializer
Added Match the Empty Query test case
Added Disjunctive Facet Filters test case and custom Deserializer


algoliasearch-common/src/main/java/com/algolia/search/inputs/query_rules/ConsequenceParams.java
```diff
@@ -129,12 +140,12 @@ public Long getMaxFacetHits() {
...
 
-  public ConsequenceParams setMaxFacetHits(Integer maxFacetHits) {
+  @JsonSetter
+  public ConsequenceParams setMaxFacetHits(Long maxFacetHits) {
     return (ConsequenceParams) super.setMaxFacetHits(maxFacetHits);
   }
 
-  @JsonSetter
-  public ConsequenceParams setMaxFacetHits(Long maxFacetHits) {
+  public ConsequenceParams setMaxFacetHits(Integer maxFacetHits) {
```
ElPicador, line 148: why not add a deprecation warning on those setters with an `Integer` ?

algoliasearch-common/src/main/java/com/algolia/search/inputs/query_rules/ConsequenceParams.java
```diff
@@ -473,13 +484,47 @@ public ConsequenceParams setIgnorePlurals(IgnorePlurals ignorePlurals) {
...
+            super.setDisableTypoToleranceOnAttributes(disableTypoToleranceOnAttributes);
   }
 
   @JsonAnySetter
   public ConsequenceParams addCustomParameter(String key, String value) {
     return (ConsequenceParams) super.addCustomParameter(key, value);
   }
 }
+
+class AutomaticFacetFilterDeserializer extends JsonDeserializer {
```
ElPicador, line 501: 👍

algoliasearch-common/src/main/java/com/algolia/search/objects/EditType.java
```diff
@@ -0,0 +1,18 @@
+package com.algolia.search.objects;
+
+
+public enum EditType {
```
ElPicador: I wouldn't use `enum`. I agree it's better for the DX, but it's more work to maintain and it needs a release of a new version of the API client if we need to add one value. I would stick to `String`

ElPicador (commented):
Nice PR 👍 
My comments are mostly nits.
Lots of tests ❤️

algoliasearch-common/src/main/java/com/algolia/search/inputs/query_rules/ConsequenceParams.java
aseure: Right, let's deprecate the useless one indeed.

algoliasearch-common/src/main/java/com/algolia/search/objects/TimeRange.java
```diff
@@ -0,0 +1,34 @@
+package com.algolia.search.objects;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import java.io.Serializable;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class TimeRange implements Serializable {
+
+  private long from;
+  private long until;
```
aseure: I know it's a bit longer to implement, but from a user PoV, I guess it would be easier if we were to provide [Java 8 `Time` instances](https://www.baeldung.com/java-8-date-time-intro) rather than `long`s. I've already written `CustomZonedDateTimeSerializer`/`CustomZonedDateTimeDeserializer` for the `ABTest.java` that you can probably reuse.

algoliasearch-common/src/main/java/com/algolia/search/inputs/query_rules/ConsequenceParams.java
Ant-hem: Yes! Will do.

algoliasearch-common/src/main/java/com/algolia/search/objects/EditType.java
Ant-hem: Ok! I will drop it.

algoliasearch-common/src/main/java/com/algolia/search/objects/TimeRange.java
Ant-hem: Ok! nice I'll update it.

PR #511, opened 2018-10-15
Fix: objectID is required to deleteObject

Ant-hem:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change
`index.deleteObject` needs a non-empty/null objectID

If the objectID is empty, the entire index will be deleted instead of one record.

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
@@ -807,6 +807,12 @@ public class APIClient {
 
   Task deleteObject(String indexName, String objectID, RequestOptions requestOptions)
       throws AlgoliaException {
+
+    if (objectID.trim().length() == 0)
+    {
+      throw new AlgoliaException("ObjectID must not be empty");
+    }
+
     Task result =
         httpClient.requestWithRetry(
             new AlgoliaRequest<>(
--- a/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
@@ -808,9 +808,8 @@ public class APIClient {
   Task deleteObject(String indexName, String objectID, RequestOptions requestOptions)
       throws AlgoliaException {
 
-    if (objectID.trim().length() == 0)
-    {
-      throw new AlgoliaException("ObjectID must not be empty");
+    if (objectID.trim().length() == 0) {
+      throw new IllegalArgumentException("ObjectID must not be empty");
     }
 
     Task result =
--- a/algoliasearch-common/src/main/java/com/algolia/search/AsyncObjects.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/AsyncObjects.java
@@ -1,6 +1,5 @@
 package com.algolia.search;
 
-import com.algolia.search.exceptions.AlgoliaException;
 import com.algolia.search.objects.RequestOptions;
 import com.algolia.search.objects.tasks.async.AsyncTask;
 import com.algolia.search.objects.tasks.async.AsyncTaskIndexing;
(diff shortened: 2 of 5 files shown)


algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -807,6 +807,12 @@ private TaskSingleIndex batchSingleIndex(
 
   Task deleteObject(String indexName, String objectID, RequestOptions requestOptions)
       throws AlgoliaException {
+
+    if (objectID.trim().length() == 0)
+    {
+      throw new AlgoliaException("ObjectID must not be empty");
```
aseure: nit: What do you think of throwing an instance of a subclass of `AlgoliaException`? Same reason as usual, from the user POV it may be easier to catch this specific error.

algoliasearch-common/src/main/java/com/algolia/search/AsyncObjects.java
```diff
@@ -217,8 +218,8 @@
    * @param requestOptions Options to pass to this request
    * @return the associated AsyncTask
    */
-  default CompletableFuture<AsyncTask> deleteObject(
-      @Nonnull String objectID, @Nonnull RequestOptions requestOptions) {
+  default CompletableFuture<AsyncTask> deleteObject (
```
aseure: Watch out of the extra trailing space

aseure (changes requested):

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
Ant-hem: @aseure Could we use this built-in Java exception ?
https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -807,6 +807,12 @@ private TaskSingleIndex batchSingleIndex(
 
   Task deleteObject(String indexName, String objectID, RequestOptions requestOptions)
       throws AlgoliaException {
+
+    if (objectID.trim().length() == 0)
+    {
```
ElPicador: 😱 
We are using google formatted for java. You can run `mvn com.coveo:fmt-maven-plugin:format` to format the code automatically

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
ElPicador: build in java is nice 👍

algoliasearch-common/src/main/java/com/algolia/search/AsyncObjects.java
```diff
@@ -206,7 +207,7 @@
    * @param objectID the unique identifier of the object to retrieve
    * @return the associated AsyncTask
    */
-  default CompletableFuture<AsyncTask> deleteObject(@Nonnull String objectID) {
+  default CompletableFuture<AsyncTask> deleteObject(@Nonnull String objectID) throws AlgoliaException {
```
ElPicador: careful, you are changing the signature of the method, it's not backward compatible

ElPicador (changes requested):

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
Ant-hem: Yup! Will fix it.

algoliasearch-common/src/main/java/com/algolia/search/AsyncObjects.java
Ant-hem: Ok! you're right it's wiser to remove it so.

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
aseure: 👍

aseure (approved):

Ant-hem:
@ElPicador I've made some changes. If you want to check! 
Thanks!

ElPicador (approved):

PR #517, opened 2018-10-30
changed: Use getters/setters of base class Query for ConsequenceParams

aseure:
The current `ConsequenceParams` class lets the user build instances
using a builder pattern. This way, a user can do:

```
ConsequenceParams params = new ConsequenceParams()
                                 .setFilters(...)
                                 .setAutomaticFacetFilters(...)
```

Because `ConsequenceParams` can accept all the possible parameters of a
`Query` object, it extends it. However, because the `ConsequenceParams`
is expected as a return type of each setter, we couldn't simply expect
to use the setters of the `ConsequenceParams` base class `Query`, even
though they are exactly the same.

This commit proposes to use a Java technique called self-referencing
generics to replace the concrete return type of the `Query` base class
setters with a self-referencing `T` generic type. Using this, all
setters can now be shared between the base class and the derived class,
with a the appropriate type signature, without having to override them
in the derived class.

To make it more clear, here is an minimal example of the code we had
before:

```
class A {
  protected int i;

  public A setI(int i) {
    this.i = i;
    return this;
  }
}

class B<T> extends A {
  private T something;

  public B setI(int i) {
    this.i = i;
    return this;
  }

  public B setSomething(T something) {
    this.something = something;
    return this;
  }
}
```

vs. the code we have now:

```
class A<T extends A<T>> {
  protected int i;

  public T setI(int i) {
    this.i = i;
    return (T) this;
  }
}

class B<T> extends A<B<T>> {
  private T something;

  public B setSomething(T something) {
    this.something = something;
    return this;
  }
}
```

where `A`/`B` respectively map to `Query`/`ConsequenceParams`. As shown,
all the setters are completely made unnecessary in the derived class.


aseure:
Thanks for taking the time to help on this @PLNech @q-litzler 

Another thank you to @nitnelave for telling me about [CRTP](https://en.m.wikipedia.org/wiki/Curiously_recurring_template_pattern). This pattern is already well known in C++ and is called polymorphic chaining. The technique is exactly the same.

algoliasearch-common/src/main/java/com/algolia/search/objects/Query.java
```diff
@@ -12,7 +12,7 @@
 @SuppressWarnings({"unused", "WeakerAccess"})
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class Query implements Serializable {
+public class Query<T extends Query<?>> implements Serializable {
```
Ant-hem: @aseure Maybe could you add a comment on that line? Just to explain what this «trick» is exactly doing. I think this is not trivial for everyone.

algoliasearch-tests/src/test/java/com/algolia/search/SyncAlgoliaIntegrationTest.java
```diff
@@ -74,7 +74,7 @@ protected Rule generateRule(String objectID) {
     Consequence ruleConsequence =
         new Consequence()
             .setUserData(ImmutableMap.of("a", "b"))
-            .setParams(new ConsequenceParams().setFacets("a=1"));
+            .setParams(new ConsequenceParams<>().setFacets("a=1"));
```
Ant-hem, line 77: Should we add a comment here as well to explain the trick?

algoliasearch-common/src/main/java/com/algolia/search/objects/Query.java
nitnelave: Maybe a name a bit more expressive than T would be good too, like "Derived". And I don't know if it's already the case, but you should hide the constructor of Query so no one can build it directly (doesn't play well with CRTP)

algoliasearch-tests/src/test/java/com/algolia/search/SyncAlgoliaIntegrationTest.java
nitnelave: I would be careful with this code: since you're using the generic class with type argument, you're accessing the raw class (the type-erased one). It's not clear that you're doing that, or even that it works. I would put the type argument for clarity's sake

algoliasearch-common/src/main/java/com/algolia/search/objects/Query.java
aseure: @Ant-hem Will add a comment for sure.

@nitnelave Sure, `T` is probably not the best name. Old habits die hard. Regarding the `Query` constructor, we need to keep it since both class can be instantiated in different contexts (actually, the `Query` class is the most used one). What makes you think it's a bad idea? With C++, I'd guess the instantiated would become more complicated with instances of the base class but with Java, it seems that instantiation remains the same.

algoliasearch-tests/src/test/java/com/algolia/search/SyncAlgoliaIntegrationTest.java
aseure: @Ant-hem Yes good point, I'll add it as well.

@nitnelave In that case, the parametric type is useless. It is only used when some specific setters are called and set the internal `T` members. As we don't play with those members, I guess the inferred type is just `Object` and it's OK in our case.

algoliasearch-common/src/main/java/com/algolia/search/objects/Query.java
nitnelave: The problem is that if you use Query directly, you're using the raw type, which is basically a Java hack to let you write `List<A> a = new List<>();`. It's not really supposed to be used as a type. What I would do if you need to have a query class is move the CRTP and setters to an abstract class (e.g. QueryImpl) and have query just be an extension of that class with the right params.

Plus in terms of design, inheriting from a concrete class is usually bad practice.

algoliasearch-tests/src/test/java/com/algolia/search/SyncAlgoliaIntegrationTest.java
nitnelave: The problem is that the type argument is not inferred, it's erased, and you're accessing the raw type (see http://www.javapractices.com/topic/TopicAction.do?Id=224 for instance). It's a backward compatibility hack that this even compiles. It would be just like write `std:: vector v;` in C++

algoliasearch-common/src/main/java/com/algolia/search/objects/Query.java
aseure: You're right. I'll add a common abstract class from which both `Query` and `ConsequenceParams` inherit. Thanks.

ElPicador (approved):

algoliasearch-tests/src/test/java/com/algolia/search/SyncAlgoliaIntegrationTest.java
aseure: After discussing it internally, we agreed that the most pragmatic choice was to let the diamond operation anyway, except for the places where `setAutomaticFacetFilters` and `setAutomaticOptionalFacetFilters` are called.

algoliasearch-common/src/main/java/com/algolia/search/objects/Query.java
aseure: Done in second commit.

aseure:
@Ant-hem I've added the commit to introduce the abstract base class `QueryBase` from which both `Query` and `ConsequenceParams` now inherit. I've also added a comment to `QueryBase` to explain the special typing of the class and why we did so. Let me know if it's OK by you now.

Thanks again for your help and suggestions @nitnelave. Really appreciate it. :)

Ant-hem:
> @Ant-hem I've added the commit to introduce the abstract base class `QueryBase` from which both `Query` and `ConsequenceParams` now inherit. I've also added a comment to `QueryBase` to explain the special typing of the class and why we did so. Let me know if it's OK by you now.
> 
> Thanks again for your help and suggestions @nitnelave. Really appreciate it. :)

Seems very good! Nice work! 👍

Ant-hem (approved):

PR #524, opened 2018-12-10
Introducing Insights/Perso

Ant-hem:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | yes

Introducing the insights feature
 
``` java

SyncUserInsightsClient insights = new SyncInsightsClient("applicationId", "apiKey").user("userToken");

// Click events (you can pass request options to every call
insights.clickedFilters("eventName", "indexName", []string{"filter"});
insights.clickedObjectIDs("eventName", "indexName", []string{"objectID"});
insights.clickedObjectIDsAfterSearch("eventName", "indexName", []string{"objectID"}, []string{42} , "queryID");

// Conversion events
insights.convertedObjectIDs("eventName", "indexName", []string{"objectID"});
insights.convertedObjectIDsAfterSearch("eventName", "indexName", []string{"objectID"}, "queryID");

// View events
insights.viewedFilters("eventName", "indexName", []string{"filter"});
insights.viewedObjectIDs("eventName", "indexName", []string{"objectID"});
```

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
@@ -11,13 +11,15 @@ import com.algolia.search.inputs.*;
 import com.algolia.search.inputs.analytics.ABTest;
 import com.algolia.search.inputs.batch.*;
 import com.algolia.search.inputs.partial_update.PartialUpdateOperation;
-import com.algolia.search.inputs.personalization.PersonalizationStrategyRequest;
+import com.algolia.search.inputs.personalization.StrategyRequest;
 import com.algolia.search.inputs.query_rules.Rule;
 import com.algolia.search.inputs.synonym.AbstractSynonym;
 import com.algolia.search.iterators.IndexIterable;
 import com.algolia.search.objects.*;
 import com.algolia.search.objects.tasks.sync.*;
 import com.algolia.search.responses.*;
+import com.algolia.search.responses.personalization.SetStrategyResult;
+import com.algolia.search.responses.personalization.GetStrategyResult;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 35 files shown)


Ant-hem:
Working on async insight tomorrow.

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
```diff
@@ -1750,6 +1751,42 @@ public SearchUserIDs searchUserIDs(
...
+  public PersonalizationStrategyResult getStrategy(@Nonnull RequestOptions requestOptions)
+      throws AlgoliaException {
+
+    return httpClient.requestWithRetry(
+        new AlgoliaRequest<>(
+            HttpMethod.GET,
+            AlgoliaRequestKind.SEARCH_API_READ,
+            Arrays.asList("1", "recommendation", "personalization", "strategy"),
+            requestOptions,
+            PersonalizationStrategyResult.class));
```
aseure: I don't know how the strategy features of Personalization will evolve but I'm not sure this name is convenient as a return type of `getStrategy`. Same remark for `setStrategy` response. What about:

* `PersonalizationStrategyResult -> GetStrategyResult` for `getStrategy` response
* `PersonalizationSaveStrategyResult -> SetStrategyResult` for `setStrategy` response

?

algoliasearch-common/src/main/java/com/algolia/search/AsyncInsightsClient.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /**
+   * @param config InsightsClient
+   * @param client APIClient
+   */
+  public AsyncInsightsClient(@Nonnull InsightsConfig config, @Nonnull AsyncAPIClient client) {
+    this.config = config;
+    this.client = client;
+    this.host = "insights." + config.getRegion() + ".algolia.io";
```
aseure, line 56: Is `getRegion` only used to set the Insights region or for Analytics as well? If it's only for Insights, could we rename the parameter and its getter `getInsightsRegion` to ensure we are talking about the Insights API. Also, if it's used for both Analytics and Insights API, I'm wondering if it makes sense to decorrelate them (probably not). What do you think @nunomaduro @julienbourdeau?

algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
Ant-hem: Yes, make sense. I will rename it. I have the same name in C# should I change it as well?

Ant-hem: Fixed with 947d1d1 @aseure

PR #527, opened 2018-12-17
make QueryBase.facets a List<String> and update tests accordingly

### LINKED ISSUE #526: Facet documentation is incorrect
- Algolia Client Version: 2.21.0
- Language Version: 1.8.0_162

### Description
The documentation for faceting is incorrect, however I think the actual problem lies in the implementation. [This](https://www.algolia.com/doc/api-reference/api-parameters/facets/) states that you can supply facets as an Array, which is what the Algolia backend seems to expect (eg, `facets=["foo","bar"]`). 

However, `QueryBase` defines facets as being a String, which when serialized will produce `facets="foo"`.

This also breaks seamless integration with the instant-search library, as it sends facets as a JSON Array to whatever backend is specified. If we use this library as our backend it will fail when attempting to facet. One must intercept the output of react-instantsearch before the query is submitted to the backend and modify the facets parameter so it is a String.

pjjanak:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #526 
| Need Doc update   | no

## Describe your change

As per #526 and per the document for [Facets](https://www.algolia.com/doc/api-reference/api-parameters/facets/), make BaseQuery.facets a List<String>.

See #526.

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/APIClient.java
@@ -18,8 +18,8 @@ import com.algolia.search.iterators.IndexIterable;
 import com.algolia.search.objects.*;
 import com.algolia.search.objects.tasks.sync.*;
 import com.algolia.search.responses.*;
-import com.algolia.search.responses.personalization.SetStrategyResult;
 import com.algolia.search.responses.personalization.GetStrategyResult;
+import com.algolia.search.responses.personalization.SetStrategyResult;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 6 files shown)


aseure:
Hello @pjjanak. Thank you for your contribution.

As answered by @Ant-hem in #526, even though your change would better reflect the current way of passing facets (according to the documentation), we cannot merge your PR as is. This is due to the compatibility break it introduces (signature changes of `getFacets`/`setFacets`). Also, those functions cannot be overloaded because of the way Jackson uses them to serialize/deserialize our objects.

If you want to pass your own facet lists to `setFacets`, I'd recommend you to [`String.join`](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.CharSequence...-).

This issue will be addressed in the upcoming major release of this API client to let you use the new type of `facets` rather than the backward-compatible one.

PR #534, opened 2019-01-15
Fixed: GeoSearch insideBounding and insidePolygon types

Ant-hem:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | yes     
| Need Doc update   | yes

## Describe your change

Replaced in the `QueryBase` class the type of `insideBoundingBox` and `insidePolygon` from `List<String>` to `List<List<Float>>`

We were not able to perfom geosearch with `List<String>`we were getting error from the API.

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/objects/QueryBase.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/objects/QueryBase.java
@@ -65,8 +65,8 @@ public abstract class QueryBase<T extends QueryBase<?>> implements Serializable 
   protected AroundRadius aroundRadius;
   protected Integer aroundPrecision;
   protected Integer minimumAroundRadius;
-  protected List<String> insideBoundingBox;
-  protected List<String> insidePolygon;
+  protected List<List<Float>> insideBoundingBox;
+  protected List<List<Float>> insidePolygon;
 
   /* highlighting-snippeting */
   protected List<String> attributesToHighlight;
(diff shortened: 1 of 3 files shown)


nunomaduro (approved):

q-litzler:
Will your `List<List<Float>>` be converted to a `String` using the format below when building the query ?

From what I tested on the dashboard, this is the expected format by the server.

```
{
  "insideBoundingBox": "46.650828100116044,7.123046875,45.17210966999772,1.009765625"
}
```

This, however, is not right.

```
{
  "insideBoundingBox": [46.650828100116044,7.123046875,45.17210966999772,1.009765625]
}
```

Ant-hem:
> Will your `List<List<Float>>` be converted to a `String` using the format below when building the query ?
> 
> From what I tested on the dashboard, this is the expected format by the server.
> 
> ```
> {
>   "insideBoundingBox": "46.650828100116044,7.123046875,45.17210966999772,1.009765625"
> }
> ```
> 
> This, however, is not right.
> 
> ```
> {
>   "insideBoundingBox": [46.650828100116044,7.123046875,45.17210966999772,1.009765625]
> }
> ```

Hey! 

I [added a test with one rectangle](https://github.com/algolia/algoliasearch-client-java-2/pull/534/files#diff-71f92d4ad97c1323493a8a6e4b87ea31R146) and the API don't reject the payload. Passing `insideBoundingBox` as a `string` is the "old way" of passing it.

aseure (approved):
Awesome. Thanks for catching this one.

q-litzler:
My mistake was: I was testing with a `List<Float>` on the dashboard, instead of a `List<List<Float>>` ! Thanks for helping me figuring this out.

PR #540, opened 2019-02-27
Introducing: Java client v3

Ant-hem:
# Introducing Java Client v3 ☕️

# ChangeLog/ New features

## Maven

To get the beta version add this to your pom.xml

```xml
<dependency>
         <groupId>com.algolia</groupId>
        <artifactId>algoliasearch-core</artifactId>
        <version>3.0.0-beta-2</version>
 </dependency>

<dependency>
        <groupId>com.algolia</groupId>
         <artifactId>algoliasearch-apache</artifactId>
         <version>3.0.0-beta-2</version>
</dependency>
```

## Less dependencies!
- org.apache.httpcomponents.httpasyncclient
- com.fasterxml.jackson.core / jackson-core / jackson-databind /jackson-datatype-jsr310
- And that's it...

## Injectable HttpCient
The default HttpClient is httpasyncclient from Apache. If you want another one, just implement the `
IHttpRequester` interface with your favorite HttpClient and inject it!

## New test suite!
- Powered by JUnit5
- Running in less than 30 seconds locally! ⚡️

## Asynchronous and synchronous methods in the same class!

Before, asynchronous and synchronous methods were split in two different classes. Now, they are reunited in the same class! 

All asynchronous methods are suffixed with the `async` keyword and return a `CompletableFuture`.

```java
SearchClient searchClient = DefaultSearchClient.create(ALGOLIA_APPLICATION_ID_1, ALGOLIA_API_KEY_1);
SearchIndex<Employee> index = searchClient.initIndex("employees", Employee.class);
    
// Async Code
CompletableFuture<SearchResult<Employee>> searchFuture = index.searchAsync(new Query("algolia"));

// Sync code
SearchResult<Employee> search = index.search(new Query("algolia"))
```
## Useful helper: waiTask()

All Algolia's asynchronous operations can now be awaited like this:

```java
index.setSettings(indexSettings).waitTask();
```

## Static factories

We added static factories to Algolia's POJOs such as `Synonyms`, `Rules` to help you build valid Algolia's objects.

```java
    Synonym wiiToWiiu =
        Synonym.createOneWaySynonym("wii_to_wii_u", "wii", Collections.singletonList("wii u"));
```

# Getting started locally:
- Clone the repo and checkout on the `develop` branch.

- Build the project
```sh
mvn clean install
```

- Run the tests
```sh
mvn clean test
```


algoliasearch/pom.xml
```diff
@@ -25,6 +25,11 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.asynchttpclient</groupId>
+      <artifactId>async-http-client</artifactId>
+      <version>2.7.0</version>
```
ElPicador: it's a good practice to put the version in a property, at the beginning of the file

algoliasearch/src/main/java/com/algolia/search/clients/AlgoliaConfig.java
```diff
@@ -0,0 +1,121 @@
...
+            put("Accept", "application/json");
+          }
+        };
+  }
+
+  public String getApplicationID() {
+    return applicationID;
+  }
+
+  public AlgoliaConfig setApplicationID(String applicationID) {
```
ElPicador: 😢, mutability

Ant-hem: Good catch! I shouldn't be too confident with intelij generators 😅

algoliasearch/src/main/java/com/algolia/search/clients/InsightsConfig.java
```diff
@@ -0,0 +1,31 @@
...
+import java.time.ZoneOffset;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.List;
+
+@SuppressWarnings("WeakerAccess")
+public class InsightsConfig extends AlgoliaConfig {
+
+  public InsightsConfig(String applicationID, String apiKey) {
+    this(applicationID, apiKey, "us");
```
ElPicador: Only `us`?

algoliasearch/src/main/java/com/algolia/search/clients/AnalyticsClient.java
```diff
@@ -0,0 +1,314 @@
...
+   * all search requests. Associated metadata and metrics are still stored
+   *
+   * @param id The ABTest ID
+   * @param requestOptions Options to pass to this request
+   */
+  public CompletableFuture<StopAbTestResponse> stopABTestAsync(
+      long id, RequestOptions requestOptions) {
+    return transport.executeRequestAsync(
+        HttpMethod.POST,
+        "/2/abtests/" + id + "/stop",
```
ElPicador: I would make sure `id` is properly escaped with a URL encoding. Same comment for everytime we are building URLs

algoliasearch/src/main/java/com/algolia/search/clients/InsightsConfig.java
ElPicador: only `US`?

algoliasearch/src/main/java/com/algolia/search/clients/AnalyticsClient.java
julienbourdeau: I'd recommend something similar to what I did in PHP. I defined a helper method to escape all variables. It looks like a sprintf call.

```php
api_path("/2/abtests/%s/stop", id)
api_path('/1/indexes/%s/facets/%s/query', indexName, facetName),
```

https://github.com/algolia/algoliasearch-client-php/blob/master/src/Support/Helpers.php#L26-L35

Ant-hem: Thanks 👍 I have did the same in .NET client. I will add it there.

algoliasearch/src/main/java/com/algolia/search/clients/InsightsConfig.java
Ant-hem: `US` is the default one at the moment isn't it? 
That's why we provide another constructor to override it, if it's not the good one.

algoliasearch/src/main/java/com/algolia/search/helpers/QueryStringHelper.java
```diff
@@ -0,0 +1,25 @@
...
+
+  public static String urlEncodeUTF8(String s) {
+    try {
+      return URLEncoder.encode(s, "UTF-8");
+    } catch (UnsupportedEncodingException e) {
+      throw new UnsupportedOperationException(e);
+    }
+  }
+
+  public static String buildQueryString(Map<String, String> map) {
```
ElPicador: no escaping of values?

algoliasearch/src/main/java/com/algolia/search/clients/AnalyticsClient.java
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  public AnalyticsClient(@Nonnull AnalyticsConfig config) {
+    this(config, new AlgoliaHttpRequester(config));
+  }
+
+  public AnalyticsClient(@Nonnull AnalyticsConfig config, @Nonnull IHttpRequester httpRequester) {
+
+    Objects.requireNonNull(httpRequester, "An httpRequester is required.");
+    Objects.requireNonNull(config, "A configuration is required.");
```
ElPicador: No needed as the parameters are already marked as @Nonnull

algoliasearch/src/main/java/com/algolia/search/clients/AlgoliaConfig.java
```diff
@@ -0,0 +1,110 @@
...
+
+  public AlgoliaConfig(String applicationID, String apiKey, ExecutorService executor) {
+    this.applicationID = applicationID;
+    this.apiKey = apiKey;
+    this.executor = executor;
+    this.batchSize = 1000;
+    this.defaultHeaders =
+        new HashMap<String, String>() {
+          {
+            put("X-Algolia-Application-Id", getApplicationID());
```
ElPicador: I would use constants for constant strings

algoliasearch/src/main/java/com/algolia/search/clients/AnalyticsConfig.java
```diff
@@ -0,0 +1,27 @@
...
+@SuppressWarnings("WeakerAccess")
+public class AnalyticsConfig extends AlgoliaConfig {
+
+  public AnalyticsConfig(String applicationID, String apiKey) {
+    super(applicationID, apiKey);
+
+    List<StatefulHost> hosts =
+        Collections.singletonList(
+            new StatefulHost(
+                "analytics.algolia.com",
```
ElPicador: I would use the regional endpoint. analytics.us.algolia.com

As we have more than one

algoliasearch/src/main/java/com/algolia/search/clients/InsightsConfig.java
ElPicador: 👌

ElPicador (commented):
Really nice improvements 👍

algoliasearch/src/main/java/com/algolia/search/clients/AnalyticsClient.java
Ant-hem: `@NonNull` is only an annotation and won't throw an exception at runtime. I think it's better to have the double check to fail fast. IMO the combination of both is complementary.

algoliasearch-common/src/main/java/com/algolia/search/Defaults.java
```diff
@@ -19,7 +20,10 @@
 
   ObjectMapper DEFAULT_OBJECT_MAPPER =
       new ObjectMapper()
+          .registerModule(new JavaTimeModule())
           .enable(OBJECT_MAPPER_DEFAULT_FEATURE)
+          .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+          .enable(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
```
redox: Could you add some comments/docs here, why enabling & disabling those?

Ant-hem: Fixed with eeb6b15! TLDR;

`DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY`: Enabled to allow Jackson to deserialize List/Array of one element.

`SerializationFeature.WRITE_DATES_AS_TIMESTAMPS`: Disabled because most of the dates we **send** to the API are ISO String.

However, those can be overridden with annotations!

redox: > `SerializationFeature.WRITE_DATES_AS_TIMESTAMPS`: Disabled because most of the dates we **send** to the API are ISO String.

Since JSON doesn't support date, we should always expect either a string (but we should not do any parsing whatsovever) or an integer I think.

redox: So, OK for converting them to timestamp I think

Ant-hem:
ping @BenoitPerrot ! :) Integrated some of the improvements we discussed yesterday.

.run-travis.sh
```diff
@@ -4,10 +4,10 @@ set -e
 
 if [[ "${JAVA_VERSION}" = "8" ]]; then
   if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [[ ! "$TRAVIS_PULL_REQUEST_SLUG" =~ ^algolia\/ ]]; then
-    eval $(./algolia-keys export) && mvn clean test jacoco:report jacoco:report-aggregate -B && mvn -pl algoliasearch-tests coveralls:report -B;
```
PLNech, line 7: Did you choose to stop reporting Code coverage?

pom.xml
```diff
@@ -1,30 +1,23 @@
...
   <description>Java client for Algolia Search API</description>
   <url>https://github.com/algolia/algoliasearch-client-java-2</url>
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-
-    <slf4j.version>1.7.25</slf4j.version>
-    <apache-http-client.version>4.5.5</apache-http-client.version>
-    <guava.version>24.0-jre</guava.version>
+    <async-http-client.version>4.1.4</async-http-client.version>
```
PLNech: Why the regression from `4.5.5` to `4.1.4`? :thinking:

.run-travis.sh
PLNech: Still `jacoco` in `pom.xml`

src/main/java/com/algolia/search/SearchClient.java
```diff
@@ -0,0 +1,197 @@
...
+   *
+   * @param indexName The name of the Algolia index
+   * @throws NullPointerException When indexName is null or empty
+   */
+  public SearchIndex<?> initIndex(@Nonnull String indexName) {
+
+    Objects.requireNonNull(indexName, "indexName can't be null.");
+
+    if (AlgoliaUtils.isEmptyWhiteSpace(indexName)) {
+      throw new NullPointerException("The index name is required");
```
PLNech: On these cases I would throw an `InvalidArgumentException`, as the test does not fail only when the String is null

src/main/java/com/algolia/search/util/AlgoliaUtils.java
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Ensure that the objectID field or the @JsonProperty(\"objectID\")" is present in the given
+   * class
+   *
+   * @param clazz The class to scan
+   * @throws AlgoliaRuntimeException When the class doesn't have an objectID field or a Jackson
+   *     annotation @JsonProperty(\"objectID\"")
+   */
+  public static <T> void ensureObjectID(@Nonnull Class<T> clazz) {
```
PLNech: Seems to duplicate the logic of `getObjectID`, which could be leveraged here :upside_down_face:

src/test/java/com/algolia/search/integration/insights/InsightsTest.java
```diff
@@ -0,0 +1,77 @@
...
+  protected final SearchClient searchClient;
+  protected final InsightsClient insightsClient;
+
+  protected InsightsTest(SearchClient searchClient, InsightsClient insightsClient) {
+    this.searchClient = searchClient;
+    this.insightsClient = insightsClient;
+  }
+
+  @Test
+  void testInsights() throws ExecutionException, InterruptedException {
```
PLNech: Nothing to assert? :thinking:

src/main/java/com/algolia/search/models/HttpRequest.java
```diff
@@ -0,0 +1,82 @@
+package com.algolia.search.models;
+
+import java.io.InputStream;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+
+public class HttpRequest {
```
PLNech: Could benefit Nullability annotations (same for `HttpResponse`) :)

PLNech: Same for any model which would be manipulated by our users, when its fields are not covered by methods with Nullability annotations

src/main/java/com/algolia/search/models/apikeys/AddApiKeyResponse.java
```diff
@@ -0,0 +1,53 @@
...
+    while (true) {
+      try {
+        getApiKeyFunction.apply(getKey());
+      } catch (AlgoliaApiException ex) {
+        if (ex.getHttpErrorCode() == 404) {
+          try {
+            Thread.sleep(1001);
+          } catch (InterruptedException ignored) {
+          }
+          continue;
```
PLNech: Isn't it a problem to ignore a `InterruptedException` and simply `continue`? :thinking: 

- https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2142
- https://stackoverflow.com/questions/28544867/is-it-ok-to-ignore-interruptedexception-if-nobody-calls-interrupt/28545020#28545020

PLNech (commented):
Superficial review considering the changeset's size, but still a few (hopefully useful) comments :upside_down_face:

pom.xml
Ant-hem: Because it's not the same `HttpClient` 😄 

- https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient (old one sync http client)
- https://mvnrepository.com/artifact/org.apache.httpcomponents/httpasyncclient (new one async http client)

src/main/java/com/algolia/search/SearchClient.java
Ant-hem: 👍 I was not sure about this one. thanks

src/main/java/com/algolia/search/models/apikeys/AddApiKeyResponse.java
Ant-hem: 👍 Good catch I'll have a look at it.

.run-travis.sh
Ant-hem: Nop I don't want to drop it! I removed it temporarily at the begining of the project. I'll add it back in a few days :)

src/main/java/com/algolia/search/util/AlgoliaUtils.java
Ant-hem: I'll have a look

src/test/java/com/algolia/search/integration/insights/InsightsTest.java
Ant-hem: It's how we wrote the test for the CTS. The insights API is mostly fire&forget API if we don't get a `400` from the API it's OK because we can't retrieve what was sent to check if it's the same on the server side (as we do for settings for example)

src/main/java/com/algolia/search/models/HttpRequest.java
Ant-hem: Could you please elaborate? 🤔 Do you mean to add annotations on the constructors parameters?

PLNech: I mean adding annotations wherever our users might benefit it:
- Constructor parameters of classes they could instantiate (not convinced it would be relevant here, as I don't think they have to create a `new HTTPRequest()` anytime)
- Annotations on public fields / public {g,s}etters that the users would manipulate to avoid unnecessary null-checks (e.g. if `headers` is never `null`, would mark `getHeaders()` as `@NonNull`) 

I think this can be applied more consistently in some classes, e.g. :
- `new Variant(...)` takes a `@NonNull String index` in constructor
-> `Variant.getIndex()` could be annotated as `@NonNull`
-> `Variant.setIndex()` should have its param annotated as `@NonNull`

You might want to use Intellij's `Infer Nullity` to quickly generate consistent annotations based on your current codebase :upside_down_face:

pom.xml
PLNech: Haha the devil is in the details :upside_down_face:

src/main/java/com/algolia/search/models/HttpRequest.java
Ant-hem: Thanks for the clarification! Neat idea. I'll look at it.

src/main/java/com/algolia/search/models/apikeys/AddApiKeyResponse.java
Ant-hem: Fixed with 049d954 thanks for catching it

src/main/java/com/algolia/search/SearchClient.java
Ant-hem: Fixed with 8bf447d

PLNech (approved):
Provided you resolve the last unsolved discussions, LGTM :ok_hand:

Ant-hem:
> Provided you resolve the last unsolved discussions, LGTM 👌

Thanks! Still need to have a look at the Nullability annotations for some POJOs :)

PR #542, opened 2019-04-04
layout: Merge "main" and "test" modules

BenoitPerrot:
To follow the most conventional structure in the Java world.

| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | None
| Need Doc update   | no

## Describe your change

Move algoliasearch/src/main and algoliasearch-test/src/test at the root level (in src/), effectively moving away from a multi package (one package for main code, another one for tests) to a single package.  

The diff is big due to files being moved (if you know a way of making that clear in Github, please share). The relevant differences are on the two pom.xml being removed, their payload being put into the "parent" pom.xml. 

The proposed structure (both main and tests in the same project) is the most conventional one in the Java world.


Ant-hem (approved):
LGTM. I plan to merge it this morning so I can move forward on others refactoring points.

PR #549, opened 2019-04-08
Added: Configuration Builder

Ant-hem:
## Describe your change
This PR is adding a builder pattern to the configuration of the clients.

Mutability of the current configuration.

### DIFF
--- a/src/main/java/com/algolia/search/AnalyticsConfig.java
+++ b/src/main/java/com/algolia/search/AnalyticsConfig.java
@@ -16,13 +16,7 @@ public final class AnalyticsConfig extends ConfigBase {
      * @param apiKey The API Key
      */
     public Builder(@Nonnull String applicationID, @Nonnull String apiKey) {
-      super(applicationID, apiKey);
-
-      List<StatefulHost> hosts =
-          Collections.singletonList(
-              new StatefulHost("analytics.algolia.com", EnumSet.of(CallType.READ, CallType.WRITE)));
-
-      this.setDefaultHosts(hosts);
+      super(applicationID, apiKey, createHosts(applicationID, apiKey));
     }
 
     @Override
@@ -32,6 +26,14 @@ public final class AnalyticsConfig extends ConfigBase {
 
     public AnalyticsConfig build() {
       return new AnalyticsConfig(this);
+    }
+
+    /** Create default hosts for the analytics configuration */
+    private static List<StatefulHost> createHosts(
+        @Nonnull String applicationID, @Nonnull String apiKey) {
+
+      return Collections.singletonList(
+          new StatefulHost("analytics.algolia.com", EnumSet.of(CallType.READ, CallType.WRITE)));
     }
   }
 
--- a/src/main/java/com/algolia/search/AnalyticsConfig.java
+++ b/src/main/java/com/algolia/search/AnalyticsConfig.java
@@ -25,22 +25,6 @@ public final class AnalyticsConfig extends ConfigBase {
       this.setDefaultHosts(hosts);
     }
 
-    /**
-     * Creates a {@link AnalyticsConfig} for the search client with custom {@link StatefulHost}.
-     * Warning: Defaults hosts are not set when setting custom {@link StatefulHost}.
-     *
-     * @param applicationID The ApplicationID
-     * @param apiKey The API Key
-     * @param customHosts List of custom hosts
-     */
-    public Builder(
-        @Nonnull String applicationID,
-        @Nonnull String apiKey,
-        @Nonnull List<StatefulHost> customHosts) {
-      super(applicationID, apiKey);
-      this.setCustomHosts(customHosts);
-    }
-
     @Override
     public Builder getThis() {
       return this;
--- a/src/main/java/com/algolia/search/AnalyticsConfig.java
+++ b/src/main/java/com/algolia/search/AnalyticsConfig.java
@@ -16,7 +16,7 @@ public final class AnalyticsConfig extends ConfigBase {
      * @param apiKey The API Key
      */
     public Builder(@Nonnull String applicationID, @Nonnull String apiKey) {
-      super(applicationID, apiKey, createHosts(applicationID, apiKey));
+      super(applicationID, apiKey, createDefaultHosts());
     }
 
     @Override
@@ -29,8 +29,7 @@ public final class AnalyticsConfig extends ConfigBase {
     }
 
     /** Create default hosts for the analytics configuration */
-    private static List<StatefulHost> createHosts(
-        @Nonnull String applicationID, @Nonnull String apiKey) {
+    private static List<StatefulHost> createDefaultHosts() {
 
       return Collections.singletonList(
           new StatefulHost("analytics.algolia.com", EnumSet.of(CallType.READ, CallType.WRITE)));
--- a/src/main/java/com/algolia/search/AnalyticsConfig.java
+++ b/src/main/java/com/algolia/search/AnalyticsConfig.java
@@ -1,44 +1,57 @@
 package com.algolia.search;
 
 import com.algolia.search.models.common.CallType;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.List;
+import java.util.*;
 import javax.annotation.Nonnull;
 
-@SuppressWarnings({"WeakerAccess", "unused"})
+@SuppressWarnings({"unused"})
 public final class AnalyticsConfig extends ConfigBase {
 
-  /**
-   * Creates an {@link AnalyticsConfig} with the default host
-   *
-   * @param applicationID The Application ID
-   * @param apiKey The API Key
-   */
-  public AnalyticsConfig(String applicationID, String apiKey) {
-    super(applicationID, apiKey);
+  public static class Builder extends ConfigBase.Builder<Builder> {
 
-    List<StatefulHost> hosts =
-        Collections.singletonList(
-            new StatefulHost("analytics.algolia.com", EnumSet.of(CallType.READ, CallType.WRITE)));
+    /**
+     * Creates an {@link AnalyticsConfig} with the default host
+     *
+     * @param applicationID The Application ID
+     * @param apiKey The API Key
+     */
+    public Builder(@Nonnull String applicationID, @Nonnull String apiKey) {
+      super(applicationID, apiKey);
 
-    this.setDefaultHosts(hosts);
+      List<StatefulHost> hosts =
+          Collections.singletonList(
+              new StatefulHost("analytics.algolia.com", EnumSet.of(CallType.READ, CallType.WRITE)));
+
+      this.setDefaultHosts(hosts);
+    }
+
+    /**
+     * Creates a {@link AnalyticsConfig} for the search client with custom {@link StatefulHost}.
+     * Warning: Defaults hosts are not set when setting custom {@link StatefulHost}.
+     *
+     * @param applicationID The ApplicationID
+     * @param apiKey The API Key
+     * @param customHosts List of custom hosts
+     */
+    public Builder(
+        @Nonnull String applicationID,
+        @Nonnull String apiKey,
+        @Nonnull List<StatefulHost> customHosts) {
+      super(applicationID, apiKey);
+      this.setCustomHosts(customHosts);
+    }
+
+    @Override
+    public Builder getThis() {
+      return this;
+    }
+
+    public AnalyticsConfig build() {
+      return new AnalyticsConfig(this);
+    }
   }
 
-  /**
-   * Creates a {@link AnalyticsConfig} for the search client with custom {@link StatefulHost}.
-   * Warning: Defaults hosts are not set when setting custom {@link StatefulHost}.
-   *
-   * @param applicationID The ApplicationID
-   * @param apiKey The API Key
-   * @param customHosts List of custom hosts
-   */
-  public AnalyticsConfig(
-      @Nonnull String applicationID,
-      @Nonnull String apiKey,
-      @Nonnull List<StatefulHost> customHosts) {
-    super(applicationID, apiKey);
-
-    this.setCustomHosts(customHosts);
+  private AnalyticsConfig(Builder builder) {
+    super(builder);
   }
 }
--- a/src/main/java/com/algolia/search/ConfigBase.java
+++ b/src/main/java/com/algolia/search/ConfigBase.java
@@ -10,7 +10,7 @@ import javax.annotation.Nonnull;
 
 /** Algolia's clients common configuration */
 @SuppressWarnings("WeakerAccess")
-class ConfigBase {
+abstract class ConfigBase {
 
   private static final String javaVersion = System.getProperty("java.version");
   private static final String clientVersion = "3.0.0.0";
@@ -100,18 +100,15 @@ class ConfigBase {
       return getThis();
     }
 
-    void setDefaultHosts(List<StatefulHost> defaultHosts) {
+    T setDefaultHosts(List<StatefulHost> defaultHosts) {
       this.defaultHosts = defaultHosts;
+      return getThis();
     }
 
     /** Sets a list of specific host to target. Default hosts will be overridden. */
     public T setCustomHosts(List<StatefulHost> customHosts) {
       this.customHosts = customHosts;
       return getThis();
-    }
-
-    public ConfigBase build() {
-      return new ConfigBase(this);
     }
   }
 
--- a/src/main/java/com/algolia/search/ConfigBase.java
+++ b/src/main/java/com/algolia/search/ConfigBase.java
@@ -21,8 +21,7 @@ abstract class ConfigBase {
   private final Integer readTimeOut;
   private final Integer writeTimeOut;
   private final Integer connectTimeOut;
-  private final List<StatefulHost> defaultHosts;
-  private final List<StatefulHost> customHosts;
+  private final List<StatefulHost> hosts;
   private final ExecutorService executor;
 
   /** Config base builder to ensure the immutability of the configuration. */
@@ -35,8 +34,7 @@ abstract class ConfigBase {
     private Integer readTimeOut;
     private Integer writeTimeOut;
     private Integer connectTimeOut;
-    private List<StatefulHost> defaultHosts;
-    private List<StatefulHost> customHosts;
+    private List<StatefulHost> hosts;
     private final ExecutorService executor;
 
     /**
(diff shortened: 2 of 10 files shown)


src/main/java/com/algolia/search/AnalyticsConfig.java
```diff
@@ -1,44 +1,57 @@
...
 
-  /**
-   * Creates an {@link AnalyticsConfig} with the default host
-   *
-   * @param applicationID The Application ID
-   * @param apiKey The API Key
-   */
-  public AnalyticsConfig(String applicationID, String apiKey) {
-    super(applicationID, apiKey);
+  public static class Builder extends ConfigBase.Builder<Builder> {
```
BenoitPerrot: This shouldn't be necessary - actually isn't; see below.

src/main/java/com/algolia/search/AnalyticsConfig.java
```diff
@@ -1,44 +1,57 @@
...
+     */
+    public Builder(@Nonnull String applicationID, @Nonnull String apiKey) {
+      super(applicationID, apiKey);
 
-    this.setDefaultHosts(hosts);
+      List<StatefulHost> hosts =
+          Collections.singletonList(
+              new StatefulHost("analytics.algolia.com", EnumSet.of(CallType.READ, CallType.WRITE)));
+
+      this.setDefaultHosts(hosts);
```
BenoitPerrot: Prefer invoking "this(applicationID, apiKey, hosts)"

src/main/java/com/algolia/search/AnalyticsConfig.java
```diff
@@ -1,44 +1,57 @@
...
+    public Builder(
+        @Nonnull String applicationID,
+        @Nonnull String apiKey,
+        @Nonnull List<StatefulHost> customHosts) {
+      super(applicationID, apiKey);
+      this.setCustomHosts(customHosts);
+    }
+
+    @Override
+    public Builder getThis() {
```
BenoitPerrot: "getThis": this is super-unexpected :D

BenoitPerrot: ah damned! now completely clear. nevermind!

src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -1,44 +1,139 @@
...
-  /**
-   * Creates an default configuration with the forkJoinPool.
-   *
-   * @param applicationID The Application ID
-   * @param apiKey The API Key.
-   */
-  ConfigBase(String applicationID, String apiKey) {
-    this(applicationID, apiKey, ForkJoinPool.commonPool());
-  }
+class ConfigBase {
```
BenoitPerrot: Nice to have that now internal!
But why not keeping it abstract?

src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -1,44 +1,139 @@
...
+
+    /** Overrides the default connect timeout. Default = 2000ms In milliseconds */
+    public T setConnectTimeOut(Integer connectTimeOut) {
+      this.connectTimeOut = connectTimeOut;
+      return getThis();
+    }
+
+    void setDefaultHosts(List<StatefulHost> defaultHosts) {
+      this.defaultHosts = defaultHosts;
+    }
```
BenoitPerrot: "return getThis();"

src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -1,44 +1,139 @@
...
+
+    /** Sets a list of specific host to target. Default hosts will be overridden. */
+    public T setCustomHosts(List<StatefulHost> customHosts) {
+      this.customHosts = customHosts;
+      return getThis();
+    }
+
+    public ConfigBase build() {
```
BenoitPerrot: hmmmm wait, this "build()" shouldn't be invoked, right?

src/main/java/com/algolia/search/AnalyticsConfig.java
```diff
@@ -1,44 +1,57 @@
...
-   */
-  public AnalyticsConfig(
-      @Nonnull String applicationID,
-      @Nonnull String apiKey,
-      @Nonnull List<StatefulHost> customHosts) {
-    super(applicationID, apiKey);
-
-    this.setCustomHosts(customHosts);
+  private AnalyticsConfig(Builder builder) {
+    super(builder);
```
BenoitPerrot, line 40: The pattern I've seen the most: keeping AnalyticsConfig constructor as before (= the constructor with all parameters) but _private_ - or internal, and have the builder invokes it in its "build" method. This way the coupling between the two is loose (which is desirable from a maintenance/evolutivity point of view)

src/main/java/com/algolia/search/AnalyticsConfig.java
```diff
@@ -1,44 +1,57 @@
...
+      List<StatefulHost> hosts =
+          Collections.singletonList(
+              new StatefulHost("analytics.algolia.com", EnumSet.of(CallType.READ, CallType.WRITE)));
+
+      this.setDefaultHosts(hosts);
+    }
+
+    /**
+     * Creates a {@link AnalyticsConfig} for the search client with custom {@link StatefulHost}.
+     * Warning: Defaults hosts are not set when setting custom {@link StatefulHost}.
```
BenoitPerrot: (It would be awesome to find an API making what is happening - overriding default hosts - clear)

src/main/java/com/algolia/search/InsightsConfig.java
```diff
@@ -6,51 +6,66 @@
...
-   * @param customHosts List of custom hosts
-   */
-  public InsightsConfig(
-      @Nonnull String applicationID,
-      @Nonnull String apiKey,
-      @Nonnull List<StatefulHost> customHosts) {
-    super(applicationID, apiKey);
-
-    this.setCustomHosts(customHosts);
+  private InsightsConfig(Builder builder) {
```
BenoitPerrot, line 57: Same as above: I suggest to keep the a private InsightConfig constructor with all parameters and invoke that in "build()"

src/test/java/com/algolia/search/integration/client/TimeoutTest.java
```diff
@@ -16,10 +16,10 @@
   void testUnreachableHostExceptions() {
     try (SearchClient client =
         new SearchClient(
-            (SearchConfig)
-                new SearchConfig(ALGOLIA_APPLICATION_ID_1, ALGOLIA_API_KEY_1)
-                    .setConnectTimeOut(2)
-                    .setReadTimeOut(2))) {
+            new SearchConfig.Builder(ALGOLIA_APPLICATION_ID_1, ALGOLIA_API_KEY_1)
```
BenoitPerrot, line 19: Ah! Way better than the forced-downcasting previous version :)

src/main/java/com/algolia/search/SearchConfig.java
```diff
@@ -9,54 +9,64 @@
...
-   * @param customHosts List of custom hosts
-   */
-  public SearchConfig(
-      @Nonnull String applicationID,
-      @Nonnull String apiKey,
-      @Nonnull List<StatefulHost> customHosts) {
-    super(applicationID, apiKey);
-
-    this.setCustomHosts(customHosts);
+  private SearchConfig(Builder builder) {
```
BenoitPerrot, line 56: Same as above: I suggest to keep a private constructor with all parameters, and invoke it in "build()"

src/main/java/com/algolia/search/AnalyticsConfig.java
BenoitPerrot: Wait: why not just have a "hosts" list, defaulting to, well, default hosts, with a "setHosts" method? This would make it very clear that default hosts would get overridden.

src/main/java/com/algolia/search/ConfigBase.java
Ant-hem: My bad :( Fixed with 23c1d91

src/main/java/com/algolia/search/ConfigBase.java
Ant-hem: 👍 fixed with 23c1d91

src/main/java/com/algolia/search/ConfigBase.java
Ant-hem: Good catch! My bad again.. fixed with 23c1d91

src/test/java/com/algolia/search/integration/client/TimeoutTest.java
Ant-hem: Sure :D

src/main/java/com/algolia/search/AnalyticsConfig.java
Ant-hem: Nice! If we are on the same page, we would like to drop the default builder constructor: 

```java
    SearchConfig config =
        new SearchConfig.Builder("appID", "apiKEY", Collections.singletonList(new StatefulHost("Algolia", EnumSet.of(CallType.READ))))
            .build();
```

to the classic builder pattern. I like it :D WDYT?

```java
    SearchConfig config =
        new SearchConfig.Builder("appID", "apiKEY")
            .setCustomHosts(
                Collections.singletonList(new StatefulHost("Algolia", EnumSet.of(CallType.READ))))
            .build();
```

src/main/java/com/algolia/search/AnalyticsConfig.java
Ant-hem: Yes! I've seen both patterns, I found the current one easier to implement with the current logic of the hosts. But if we decide to drop the Builder constructor with the hosts. i.e having only one with AppID and APIKey it could be nice to have this pattern. (see above for the constructors) :)

src/main/java/com/algolia/search/AnalyticsConfig.java
Ant-hem: Or with `setHosts` instead of `setCustomHosts`

Ant-hem: Proposal: c38029d

BenoitPerrot: Yes, that's it: `setHosts`, and Builders' constructors taking only applicationID and apiKey 👍

Ant-hem: Perfect! Thanks :)

src/main/java/com/algolia/search/AnalyticsConfig.java
```diff
@@ -1,44 +1,41 @@
...
-        Collections.singletonList(
-            new StatefulHost("analytics.algolia.com", EnumSet.of(CallType.READ, CallType.WRITE)));
+    /**
+     * Creates an {@link AnalyticsConfig} with the default host
+     *
+     * @param applicationID The Application ID
+     * @param apiKey The API Key
+     */
+    public Builder(@Nonnull String applicationID, @Nonnull String apiKey) {
+      super(applicationID, apiKey);
```
BenoitPerrot: I suggest to adjust `ConfigBase`'s constructor to make it receive default hosts (rationale: it is not an optional parameter for ConfigBase):
```
super(applicationID, apiKey, Collections.singletonList(...))
```

src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -1,44 +1,136 @@
...
+
+    private final String applicationID;
+    private final String apiKey;
+    private final HashMap<String, String> defaultHeaders;
+    private int batchSize;
+    private Integer readTimeOut;
+    private Integer writeTimeOut;
+    private Integer connectTimeOut;
+    private List<StatefulHost> defaultHosts;
+    private List<StatefulHost> customHosts;
```
BenoitPerrot: I suggest to have just: `private List<StatefulHost> hosts;`. No "default" vs "custom" ones.

src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -1,44 +1,136 @@
...
+    private final ExecutorService executor;
+
+    /**
+     * Builds a base configuration
+     *
+     * @param applicationID The Algolia Application ID
+     * @param apiKey The API Key: could be Admin API Key or Search API Key
+     * @throws NullPointerException if ApplicationID/ApiKey/Config/Requester is null
+     */
+    public Builder(@Nonnull String applicationID, @Nonnull String apiKey) {
```
BenoitPerrot: Here take "defaultHosts" as a parameter, and use it to set the "hosts" attribute (the names I am using here are chosen on purpose: at construction time is given a list of default hosts)

src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -1,44 +1,136 @@
...
+      return getThis();
+    }
+
+    /** Overrides the default connect timeout. Default = 2000ms In milliseconds */
+    public T setConnectTimeOut(Integer connectTimeOut) {
+      this.connectTimeOut = connectTimeOut;
+      return getThis();
+    }
+
+    T setDefaultHosts(List<StatefulHost> defaultHosts) {
```
BenoitPerrot: (just remove this one)

src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -1,44 +1,136 @@
...
+      this.defaultHosts = defaultHosts;
+      return getThis();
+    }
+
+    /** Sets a list of specific host to target. Default hosts will be overridden. */
+    public T setHosts(List<StatefulHost> customHosts) {
+      this.customHosts = customHosts;
```
BenoitPerrot: `this.hosts = hosts`

src/main/java/com/algolia/search/InsightsConfig.java
```diff
@@ -1,56 +1,59 @@
...
+    }
+
+    /**
+     * Creates an {@link InsightsConfig} with a custom region.
+     *
+     * @param applicationID The ApplicationID
+     * @param apiKey The API Key
+     */
+    public Builder(@Nonnull String applicationID, @Nonnull String apiKey, @Nonnull String region) {
+      super(applicationID, apiKey);
```
BenoitPerrot: `super(applicationID, apiKey, Collections.singletonList(...))`

BenoitPerrot (commented):
I propose to drop `setDefaultHosts`, replacing `defaultHosts` and `customHosts` with just `hosts`

src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -1,44 +1,136 @@
...
+      this.defaultHeaders.put(
+          Defaults.USER_AGENT_HEADER,
+          String.format("Algolia for Java (%s); JVM (%s)", clientVersion, javaVersion));
+      this.defaultHeaders.put(Defaults.ACCEPT_HEADER, Defaults.APPLICATION_JSON);
+      this.defaultHeaders.put(Defaults.ACCEPT_ENCODING_HEADER, Defaults.CONTENT_ENCODING_GZIP);
+      this.executor = ForkJoinPool.commonPool();
+    }
+
+    /** To prevent unchecked cast warning. */
+    public abstract T getThis();
```
BenoitPerrot, line 67: OK, actually there is a simpler construction. Tell me what you think:

```java
abstract class ConfigBase {
  public abstract static class Builder<T> {
    // [...]
    public abstract T build();

    public Builder<T> setBatchSize(int batchSize) {
      this.batchSize = batchSize;
      return this;
    }
    // [...]
  }
}
```

And other Builders:
```java
public final class InsightsConfig extends ConfigBase {
  // [...]
  public static class Builder extends ConfigBase.Builder<InsightsConfig> {
    @Override
    public InsightsConfig build() {
      return new InsightsConfig(this);
    }
  }
  // [...]
}
```

Ant-hem:
@BenoitPerrot proposal with all your reviews 2a3538e 
Looks pretty good now :)

src/main/java/com/algolia/search/AnalyticsConfig.java
```diff
@@ -33,6 +27,14 @@ public Builder getThis() {
     public AnalyticsConfig build() {
       return new AnalyticsConfig(this);
     }
+
+    /** Create default hosts for the analytics configuration */
+    private static List<StatefulHost> createHosts(
+        @Nonnull String applicationID, @Nonnull String apiKey) {
```
BenoitPerrot: `applicationID` and `apiKey` are unused, I suggest to do without

src/main/java/com/algolia/search/InsightsConfig.java
```diff
@@ -51,6 +40,19 @@ public Builder getThis() {
     public InsightsConfig build() {
       return new InsightsConfig(this);
     }
+
+    /** Create default hosts for the insights configuration */
+    private static List<StatefulHost> createDefaultHosts(
+        @Nonnull String applicationID, @Nonnull String apiKey, @Nonnull String region) {
```
BenoitPerrot: same here: `applicationID` and `apiKey` are unused, I suggest to do without

src/main/java/com/algolia/search/SearchConfig.java
```diff
@@ -13,7 +13,21 @@
...
+      return this;
+    }
+
+    public SearchConfig build() {
+      return new SearchConfig(this);
+    }
+
+    /** Create default hosts for the search configuration */
+    private static List<StatefulHost> createDefaultHosts(
+        @Nonnull String applicationID, @Nonnull String apiKey) {
```
BenoitPerrot: here too: `applicationID` and `apiKey` are unused, I suggest to do without

BenoitPerrot (changes requested):
`applicationID` and `apiKey` are unused in createDefaultHosts methods: let's do without (if we need it in the future, we would introduce them then)

Ant-hem:
> `applicationID` and `apiKey` are unused in createDefaultHosts methods: let's do without (if we need it in the future, we would introduce them then)

Arg. Copy/paste is always bad. Good catch!

src/main/java/com/algolia/search/SearchConfig.java
Ant-hem: Fixed with eb0e845 :)

BenoitPerrot: Hmmmm are you sure? I am still seeing it in the final diff

Ant-hem: Should be better now.

Ant-hem: `applicationID` is used to create the host but `apiKey` is unused.

BenoitPerrot (approved):
Looks good!

Ant-hem:
Thank you so much for you thorough review on this PR 🎉 
We have a neat configuration builder now.

PR #550, opened 2019-04-11
test: Limit variables' scope

BenoitPerrot:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | none
| Need Doc update   | no

## Describe your change

Transform attributes of test classes into method-local variables. 

This makes the lifecycle of the properties in the tests classes more clear.

### DIFF
--- a/src/test/java/com/algolia/search/integration/client/SecuredAPIKeyTest.java
+++ b/src/test/java/com/algolia/search/integration/client/SecuredAPIKeyTest.java
@@ -10,64 +10,51 @@ import com.algolia.search.integration.models.AlgoliaObject;
 import com.algolia.search.models.apikeys.SecuredApiKeyRestriction;
 import com.algolia.search.models.indexing.BatchIndexingResponse;
 import com.algolia.search.models.indexing.Query;
-import java.io.IOException;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.Collections;
 import java.util.concurrent.CompletableFuture;
-import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @ExtendWith({IntegrationTestExtension.class})
 class SecuredAPIKeyTest {
-
-  private static SearchClient clientWithRestriction;
-  private SearchIndex<AlgoliaObject> index1;
-  private static SearchIndex<AlgoliaObject> index2;
-  private String index1Name;
-  private String index2Name;
-
-  SecuredAPIKeyTest() {
-    index1Name = getTestIndexName("secured_api_keys");
-    index2Name = getTestIndexName("secured_api_keys_dev");
-    index1 = searchClient.initIndex(index1Name, AlgoliaObject.class);
-    index2 = searchClient2.initIndex(index2Name, AlgoliaObject.class);
-  }
-
-  @AfterAll
-  static void after() throws IOException {
-    clientWithRestriction.close();
-    index2.deleteAsync();
-  }
-
   @Test
   void testSecuredAPIKey() throws Exception {
-    CompletableFuture<BatchIndexingResponse> addOneFuture =
-        index1.saveObjectAsync(new AlgoliaObject().setObjectID("one"));
-    CompletableFuture<BatchIndexingResponse> addTwoFuture =
-        index2.saveObjectAsync(new AlgoliaObject().setObjectID("one"));
+    String index1Name = getTestIndexName("secured_api_keys");
+    String index2Name = getTestIndexName("secured_api_keys_dev");
+    SearchIndex<AlgoliaObject> index1 = searchClient.initIndex(index1Name, AlgoliaObject.class);
+    SearchIndex<AlgoliaObject> index2 = searchClient2.initIndex(index2Name, AlgoliaObject.class);
+    try {
+      CompletableFuture<BatchIndexingResponse> addOneFuture =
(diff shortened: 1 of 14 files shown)


src/test/java/com/algolia/search/integration/insights/InsightsTest.java
```diff
@@ -82,5 +70,7 @@ void testInsights() throws ExecutionException, InterruptedException {
             Arrays.asList("1", "2"),
             search2.getQueryID())
         .get();
+
+    insightsClient.close();
```
Ant-hem: Maybe we could use a try with ressources to ensure `insightsClient`is closed in case of test failure?

src/test/java/com/algolia/search/integration/client/SecuredAPIKeyTest.java
```diff
@@ -69,5 +52,9 @@ void testSecuredAPIKey() throws Exception {
 
     assertThatThrownBy(() -> index2WithRestriction.search(new Query()))
         .hasMessageContaining("Index not allowed with this API key");
+
+    clientWithRestriction.close();
```
Ant-hem: Same as below :) as we don't have the `@AfterAll` anymore we could use a try with resources to close the client?

src/test/java/com/algolia/search/integration/client/SecuredAPIKeyTest.java
```diff
@@ -69,5 +52,9 @@ void testSecuredAPIKey() throws Exception {
...
+
+    index2.deleteAsync().get().waitTask();
```
Ant-hem: Should we consider putting back the `deleteAsync()` in the `@AfterAll` method? Or maybe within a `try/catch/finally` block, because If something wrong happens the index is not deleted.

Moreover, the `waitTask()` is not necessary as this deletion is a fire&forget. (Usually we perform a waitTask when we want to be sure that the task has been processed by the engine or when the next steps are dependents on this task).

Regarding the `get()`, Will the HTTP call be sent correctly if we remove it? I think so, WDYT? 🤔

BenoitPerrot: > Should we consider putting back the deleteAsync() in the @AfterAll method? Or maybe within a try/catch/finally block, because If something wrong happens the index is not deleted.

I would suggest going for a try/catch/finally block, to keep the index variable local (I realise that the index itself is global, but as much as possible I would rather keep variables locals). Will update the comment.

> `waitTask()` is not necessary as the deletion is fire&forget.

Agreed, the deletion is fire&forget, so `waitTask`, in such a test, is not necessary.

On the other hand, tests should be as illustrative of best practices as possible: I wouldn't blame a customer of ours copy/pasting a snippet of code found in tests. And if I understand correctly, we invite customers to check if not wait the result of tasks.

Let's start by staying closer to the original code, by not having `get` and `waitTask`, and see for the best-practice thing later on.

> Will the HTTP call be sent if we remove the `get()`

It depends on when the parallel execution starts - on when the future is actually triggering a new thread. I have the feeling that it happens _before_ the HTTP request is effectively done ^_^; I have to check that. Let's say that if it doesn't, it's a bug (existing in the current "AfterAll"-based code base), requiring a bug fix on its own (ie in a separated PR).

src/test/java/com/algolia/search/integration/client/SecuredAPIKeyTest.java
BenoitPerrot: Agreed. I will update the commits.

src/test/java/com/algolia/search/integration/insights/InsightsTest.java
BenoitPerrot: Agreed. When I read your comment I first had in mind to suggest putting that in another review, but it actually makes sense to have it there: prior to the patch, an error in the very test wouldn't have prevented the closing of the client, we must keep this guarantee.

src/test/java/com/algolia/search/integration/client/SecuredAPIKeyTest.java
Ant-hem: > I would suggest going for a try/catch/finally block, to keep the index variable local (I realise that the index itself is global, but as much as possible I would rather keep variables locals). Will update the comment.

👍 

> It depends on when the parallel execution starts - on when the future is actually triggering a new thread. I have the feeling that it happens _before_ the HTTP request is effectively done ^_^; I have to check that. Let's say that if it doesn't, it's a bug (existing in the current "AfterAll"-based code base), requiring a bug fix on its own (ie in a separated PR).

Let's put belt and braces and keep the `get()` 😄

src/test/java/com/algolia/search/integration/insights/InsightsTest.java
BenoitPerrot: done

src/test/java/com/algolia/search/integration/client/SecuredAPIKeyTest.java
BenoitPerrot: done

Ant-hem:
Could you please run `mvn com.coveo:fmt-maven-plugin:format` before we squash & merge this PR? 
My bad, I didn't told you before we were using this fmt plugin.

BenoitPerrot:
> Could you please run `mvn com.coveo:fmt-maven-plugin:format` before we squash & merge this PR?
> My bad, I didn't told you before we were using this fmt plugin.

No worries; I updated the commits accordingly.

Ant-hem (approved):

PR #551, opened 2019-04-11
tests: Split "core" and "apache"

BenoitPerrot:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | none
| Need Doc update   | no

## Describe your change

Extract clients from test scenarios - injecting the former in the latter

This makes injection of custom clients possible - so that the test suite can be reused for custom clients.

### DIFF
--- a/src/test/java/com/algolia/search/apache/analytics/AnalyticsTest.java
+++ b/src/test/java/com/algolia/search/apache/analytics/AnalyticsTest.java
@@ -1,0 +1,27 @@
+package com.algolia.search.apache.analytics;
+
+import static com.algolia.search.integration.IntegrationTestExtension.ALGOLIA_API_KEY_1;
+import static com.algolia.search.integration.IntegrationTestExtension.ALGOLIA_APPLICATION_ID_1;
+
+import com.algolia.search.AnalyticsClient;
+import com.algolia.search.integration.IntegrationTestExtension;
+import java.io.IOException;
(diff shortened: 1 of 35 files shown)


src/test/java/com/algolia/search/apache/analytics/AnalyticsTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static com.algolia.search.integration.IntegrationTestExtension.ALGOLIA_API_KEY_1;
+import static com.algolia.search.integration.IntegrationTestExtension.ALGOLIA_APPLICATION_ID_1;
+
+@ExtendWith({IntegrationTestExtension.class})
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class AnalyticsTest extends com.algolia.search.integration.analytics.AnalyticsTest {
+
+  AnalyticsTest() {
+    super(IntegrationTestExtension.searchClient, new AnalyticsClient(ALGOLIA_APPLICATION_ID_1, ALGOLIA_API_KEY_1));
```
Ant-hem: Here `IntegrationTestExtension.searchClient`is `null` that's why tests are failing. If I remove `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` and move `afterAll()` in the implementation I don't have the issue anymore.

PR #552, opened 2019-04-11
test: Detach "core" from "apache"

BenoitPerrot:
| Q                 | A
| Bug fix?          |no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | none
| Need Doc update   | no

## Describe your change

Move the setup of the Apache-based search client away from the package containing the actual test suite.

This actually extracts the last remaining dependency on the Apache client from the tests suite.


Ant-hem:
Re-based with fixes from `test/split-core-apache`, now tests are passing 👍

Ant-hem:
Thanks @BenoitPerrot 🎉

PR #557, opened 2019-04-15
Proposal: multi modules project

Ant-hem:
Split the library into multiple modules

`algoalisearch-core` which is httpClient agnostic
`algoliasearch-apache` which has an implementation of HttpRequester


algoliasearch-apache/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+            <groupId>com.algolia</groupId>
+            <artifactId>algoliasearch-core</artifactId>
+            <version>${algoliasearch-core-version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.algolia</groupId>
+            <artifactId>algoliasearch-core</artifactId>
+            <version>${algoliasearch-core-version}</version>
+            <type>test-jar</type>
```
BenoitPerrot, line 33: add classifier too:
```xml
<classifier>tests</classifier>
```

algoliasearch-apache/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+        <jsr305.version>3.0.2</jsr305.version>
+        <algoliasearch-core-version>3.0.0-SNAPSHOT</algoliasearch-core-version>
+    </properties>
+
+    <dependencies>
+        <!--Apache http client-->
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpasyncclient</artifactId>
+            <version>${async-http-client.version}</version>
```
BenoitPerrot: I am not sure we really need a property for this version. We do not plan to allow its overriding on the command line, and the value is used only once (here).

algoliasearch-apache/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+            <artifactId>algoliasearch-core</artifactId>
+            <version>${algoliasearch-core-version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+            <version>${jsr305.version}</version>
```
BenoitPerrot: Remove the <version> tag; the version is already specified in the dependencyManagement section of the parent pom.

algoliasearch-apache/src/main/java/com/algolia/search/apache/ApacheAnalyticsClient.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Algolia's REST analytics client that wraps an instance of the transporter. which wraps the HTTP
+ * Client This client allows to build typed requests and read typed responses. Requests are made
+ * under the Algolia's retry-strategy. This client is intended to be reused and it's thread-safe.
+ *
+ * @see <a href="https://www.algolia.com/doc/rest-api/analytics/">Algolia.com</a>
+ */
+@SuppressWarnings("WeakerAccess")
+public class ApacheAnalyticsClient {
```
BenoitPerrot: I suggest to drop the "Apache" prefix: the code is already in the com.algolia.search.**apache** package.

algoliasearch-apache/src/main/java/com/algolia/search/apache/ApacheHttpRequester.java
```diff
@@ -36,18 +42,13 @@
   private final RequestConfig requestConfig;
   private final ConfigBase config;
 
-  public ApacheHttpRequester(@Nonnull ConfigBase config) {
+  ApacheHttpRequester(@Nonnull ConfigBase config) {
```
BenoitPerrot: likewise: the Apache prefix could be dropped

algoliasearch-apache/src/main/java/com/algolia/search/apache/ApacheInsightsClient.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Algolia's REST insights client that wraps an instance of the transporter. which wraps the HTTP
+ * Client This client allows to build typed requests and read typed responses. Requests are made
+ * under the Algolia's retry-strategy. This client is intended to be reused and it's thread-safe.
+ *
+ * @see <a href="https://www.algolia.com/doc/rest-api/insights/">Algolia.com</a>
+ */
+@SuppressWarnings("WeakerAccess")
+public class ApacheInsightsClient {
```
BenoitPerrot: likewise: the Apache prefix could be dropped

algoliasearch-apache/src/main/java/com/algolia/search/apache/ApacheSearchClient.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Algolia's REST search client that wraps an instance of the transporter which wraps the HttpClient
+ * This client allows to build typed requests and read typed responses. Requests are made under the
+ * Algolia's retry-strategy. This client is intended to be reused and it's thread-safe.
+ *
+ * @see <a href="https://www.algolia.com/doc/rest-api/search/">Algolia.com</a>
+ */
+@SuppressWarnings("WeakerAccess")
+public class ApacheSearchClient {
```
BenoitPerrot: likewise: the Apache prefix could be dropped

algoliasearch-apache/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>algoliasearch-apache</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <properties>
+        <async-http-client.version>4.1.4</async-http-client.version>
+        <jsr305.version>3.0.2</jsr305.version>
```
BenoitPerrot: remove this jsr305 property (see below why)

algoliasearch-core/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+    <artifactId>algoliasearch-core</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jackson.version>2.9.8</jackson.version>
+        <slf4j.version>1.7.25</slf4j.version>
+        <jsr305.version>3.0.2</jsr305.version>
+        <assertj-core.version>3.12.1</assertj-core.version>
```
BenoitPerrot: Remove assertj-core.version: it is never used

algoliasearch-core/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+
+    <artifactId>algoliasearch-core</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jackson.version>2.9.8</jackson.version>
+        <slf4j.version>1.7.25</slf4j.version>
+        <jsr305.version>3.0.2</jsr305.version>
```
BenoitPerrot: I am not sure properties are needed for "slf4j" and "jsr305": their values are used only once each

pom.xml
```diff
@@ -1,315 +1,188 @@
...
+    <groupId>com.algolia</groupId>
+    <artifactId>algoliasearch</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <name>Algolia Search</name>
+    <description>Java client for Algolia Search API</description>
+    <url>https://github.com/algolia/algoliasearch-client-java-2</url>
+
+    <!-- sub modules -->
```
BenoitPerrot: remove this comment: the < modules > says it all :)

src/main/java/com/algolia/search/models/indexing/Query.java
```diff
@@ -521,6 +521,15 @@ public Query setEnablePersonalization(Boolean enablePersonalization) {
     return this;
   }
 
+  public Integer getPersonalizationImpact() {
```
BenoitPerrot, line 524: blargh this change is completely unrelated to the multi-module thing :) you should push that in another PR

algoliasearch-core/src/main/java/com/algolia/search/util/AlgoliaUtils.java
```diff
@@ -93,30 +89,6 @@ public static Boolean isNullOrEmptyWhiteSpace(final String stringToCheck) {
             + "must have an objectID property or a Jackson annotation @JsonProperty(\"objectID\")");
   }
 
-  public static CompletableFuture<HttpResponse> toCompletableFuture(
```
BenoitPerrot, line 96: This inlining should be done in a separate, dedicated commit: it is actually unrelated to the move to multi-module

BenoitPerrot (changes requested):
- Please address the elements written in the imperative form ("Please do that" or "Remove this")
- Feel free not to follow my suggestions on the others (eg "I am unsure we need a property" or "We could drop the prefix 'Apache' ")

pom.xml
```diff
@@ -34,6 +34,14 @@
         </developer>
     </developers>
 
+    <distributionManagement>
+        <repository>
+            <id>bintray-algolia-maven</id>
+            <name>algolia-maven</name>
+            <url>https://api.bintray.com/maven/algolia/maven/algoliasearch/;publish=1</url>
+        </repository>
+    </distributionManagement>
```
BenoitPerrot, line 50: Let's do this in another PR: it is unrelated to switching to multi-modules, and it can break things badly, so let's keep the two separate (so if we need to revert, we can revert just the use of bintray)

src/main/java/com/algolia/search/models/indexing/Query.java
Ant-hem: arg I checked out too fast 😞

algoliasearch-core/src/main/java/com/algolia/search/util/AlgoliaUtils.java
Ant-hem: It is! This method takes an `org.apache.http.HttpResponse` as an input, as we wanted  `algoliasearch-core` to be completely independent from any http implementetion we had to move this helper in  `algoliasearch-apache - ApacheRequester`.

algoliasearch-core/pom.xml
Ant-hem: I was told it's always a good practice to have all the version in properties. Should I move them?

algoliasearch-apache/pom.xml
Ant-hem: Just curious, why do we need the `classifier` tag? 

When we run `mvn deploy` we already have a `.jar` suffixed with `tests` : `algoliasearch-apache-3.0.0-beta-1-tests.jar` Am I missing something?

Ant-hem: Fixed!

algoliasearch-apache/pom.xml
Ant-hem: Fixed!

algoliasearch-apache/pom.xml
Ant-hem: I moved findbugs in the parent pom as the library is shared with both children

algoliasearch-core/pom.xml
Ant-hem: Fixed

algoliasearch-apache/src/main/java/com/algolia/search/apache/ApacheAnalyticsClient.java
Ant-hem: I don't have a strong opinion on it. Do you have example of other libraries in the Java world doing something similar? I couldn't find one.

algoliasearch-core/src/main/java/com/algolia/search/util/AlgoliaUtils.java
BenoitPerrot: > It is!

It is not: it is possible to inline this helper in ApacheHttpRequester right now - without splitting the code into multi-module. For sure, it must be done for splitting cleanly. But it can be done beforehand.

The rationale is: separate concerns, keep commits atomic - always, always, always. So that if you have to revert a commit, it's less a big deal. And so that reviewing code is less stressful on the brain.

... Up to you! (seriously)

algoliasearch-apache/src/main/java/com/algolia/search/apache/ApacheAnalyticsClient.java
BenoitPerrot: Actually we don't see it because that is the default construction. Our own classes are for example "com.algolia.search.AccountClient", and not "com.algolia.search.AlgoliaSearchAccountClient"

It's the very purpose of _namespaces_: putting things in the same named set :P

algoliasearch-core/pom.xml
BenoitPerrot: Your call. Do you remember the rationale of this best practice?

On my side I really do not see the point to keep them out-of-line if they are not to be overridden on the command line or in Maven, and if they are used only in one place.

Note that we can go like the current construction, and decide later on too.

algoliasearch-apache/src/main/java/com/algolia/search/apache/ApacheAnalyticsClient.java
Ant-hem: Why not having all the `Apache-` classes in `com.algolia.search` then it's very clear for the documentation/customers how to instantiate a client?

I find this (in doc) clearer:
```java
SearchClient client = ApacheSearchClient.create(...);
```

than: 
```java
SearchClient client = com.algolia.search.apache.SearchClient.create(....);
```

ping @aseure

Ant-hem: Or let's remane `ApacheSearchClient` by `DefaultSearchClient`

```Java
SearchClient client = DefaultSearchClient.create(...);
````

Comment (BenoitPerrot) [reply algoliasearch-apache/src/main/java/com/algolia/search/apache/ApacheAnalyticsClient.java]: This sounds like a good approach to me 👍

File: algoliasearch-apache/src/main/java/com/algolia/search/DefaultSearchClient.java [hunk 15]
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Algolia's REST search client that wraps an instance of the transporter which wraps the HttpClient
+ * This client allows to build typed requests and read typed responses. Requests are made under the
+ * Algolia's retry-strategy. This client is intended to be reused and it's thread-safe.
+ *
+ * @see <a href="https://www.algolia.com/doc/rest-api/search/">Algolia.com</a>
+ */
+@SuppressWarnings("WeakerAccess")
+public class DefaultSearchClient {
```
BenoitPerrot, line 13: (yes I definitely like it)

BenoitPerrot (approved):

Issue #556, opened 2019-04-15
Provide "uber libraries"


BenoitPerrot:
**TL;DR Let's provide, along with the current one(s), an "uber" jar library embedding all the required dependencies.**

We were mentioning lately that there is a lot of support requested by customers due to conflicting dependencies when integrating the Java client. This is especially true in the current version of the client, due to its dependency on Guava (which is a common package nowadays, especially in back-ends / in frameworks used for creating back-ends). While the new version of the client doesn't depends on Guava, the problem should be less frequent, but wouldn't entirely disappear (e.g. customers could still need conflicting versions of jackson).

The generic problem can be described as follows:
- AlgoliaSearch depends on package P in version X;
- a customer creates a product A leveraging some framework F, which depends on package P in version Y, and tries to integrate AlgoliaSearch;
- maven (or any other Java build tool actually) detects that there is something fishy: P is needed twice, in two different version X and Y;
-- best case: X and Y differ from only a minor increment, and X and Y are actually interchangeable (... and the build tool silently decides which one of X and Y win ...)
-- worst case: X and Y differ on a major increment (and we have to explain customers how to deal with it)

This problem is quite common in the Java world, and the "usage" is to _shade_. This consists in putting a particular version of a particular package in a new namespace, and updating the code using it accordingly (so, resolving the naming conflicts, but effectively embedding several versions of the same package). Concretely, this means renaming "com.google.guava" as "com.algoliasearch.google.guava" in algoliasearch.jar, and embedding the appropriate version of "com.google.guava" while renaming it. This construction is generally leveraged while building the final jar. But it can be done anytime - including when creating a jar library.

We could generate, along the "usual" jar, an "uber" jar containing all the dependencies already shaded. And explain which one to use, depending on the use case.

For example, assuming v3 comes in "algoliasearch-core" and "algoliasearch-apache":
- **You want to integrate algoliasearch with absolutely no friction? Add a dependency on "algoliasearch-apache:uber"** This package comes with the reference implementation of the HTTP requester (leveraging Apache Http client), and embeds all the required dependencies. It is bigger than the other, in exchange you don't have to manage to version conflict resolution with whatever build system, you are using
- **You want to leverage the default implementation, but want to resolve conflicts yourself? Go for "algoliasearch-apache"** This packages comes with the reference implementation of the HTTP requester, without embedding the required dependencies. It is smaller than the previous one, but you might have to deal with conflict resolution
[...]

PR #559, opened 2019-04-19
packaging: Generate uber jars for -core and -apache

BenoitPerrot:
Relocating embedded dependencies inside com.algoliasearch, filtering
out unused resources.

Note that slf4j-api is NOT embedded in the uber jars, on purpose: if
slf4j-api is embedded here, it would need to be relocated (*), yet if
gets relocated, implementations cannot be injected anymore (**),
making slf4j irrelevant.

(*) Embedding jars in an uber-library _require_ their relocation: a
library is meant to be used along with other jars, which may depend on
other versions of the jars embedded here. Relocating these mostly get
rid of conflicts; not relocating them, would on the contrary make
conflicts practically impossible to solve.

(**) the injection used by slf4j is namespace based; if the api gets
relocated, implementation would have to be relocated too.

From a "pom.xml point of view", a pluginManagement section is set in
the root pom.xml to configure the shading - but not apply it. This
configuration in then leveraged in dedicated modules
(algolia-core-uber and algolia-apache-uber) coming with their own
pom.xml files (and possibly adjusting - augmenting - the relocations,
in the apache flavor).


Ant-hem:
Ref issue: #556

Ant-hem (approved):

PR #562, opened 2019-04-23
Added: missing toString() method in models

Ant-hem:


BenoitPerrot (commented):
I am unsure of the implementation:
- couldn't reflection be leveraged to achieve ~the same?
- note that containers will probably be dumped as hashes
- ... should toString() be effectively overridden? (=> adding a field in a model will change the behavior - how would be the version number impacted :trollface: ?)

Ant-hem:
> I am unsure of the implementation:
> 
>     * couldn't reflection be leveraged to achieve ~the same?
> 
>     * note that containers will probably be dumped as hashes
> 
>     * ... should toString() be effectively overridden? (=> adding a field in a model will change the behavior - how would be the version number impacted :trollface: ?)

I added it mostly for backward compatibility with the v2. All the models in the v2 have the `toString()` overridden and I guess some of users are using it. 

Sure we could leverage reflection 👍 I took the easy way generating the `toString()` method from the IDE. IMO it's good enough for the first release.

Good point about the versioning :trollface: 

Should we move forward with this one to keep the same feature as the v2 or do we close this PR?

Ant-hem:
Let's close it for the moment and wait if we need it in the future.

PR #564, opened 2019-04-29
Added: QueryBase with CRTP to handle Query inheritance

Ant-hem:
Proposal of adding CRTP with a `QueryBase` class (we can discuss the naming) to prevent down-casting  when inheriting from `Query` instead of overriding all the methods in the children.


BenoitPerrot:
As discussed: alas in Java there is no easy way to mix-in, and we are down to leverage inheritance here. Go for it, just please rename QueryBase into SearchParameters

BenoitPerrot (approved):

PR #571, opened 2019-05-09
Make some query rule and synonym management APIs more consistent with established conventions

matthewbogner:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | no

## Describe your change
While integrating with Algolia, a couple SDK interactions seemed to not follow conventions found in other places in the existing SDK.  Namely the usage of fluent setters (that return `this`).  

Additionally deleteRule and deleteSynonym didn't have symmetry in their method signatures. 

SDK consistency

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/SyncRules.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/SyncRules.java
@@ -123,6 +123,17 @@ public interface SyncRules<T> extends SyncBaseIndex<T> {
    *
    * @param ruleId the id of the query rule
    * @param forwardToReplicas should this request be forwarded to replicas
+   * @return the associated task
+   */
+  default Task deleteRule(@Nonnull String ruleId, boolean forwardToReplicas) throws AlgoliaException {
+    return getApiClient().deleteRule(getName(), ruleId, forwardToReplicas, new RequestOptions());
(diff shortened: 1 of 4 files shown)


aseure:
Hello @matthewbogner and thank you for your contribution!

You're right, the fluent API could be better by using the correct return type for `Consequence` setters. We are gonna integrate your changes but on the `develop` branch. The reason is that the changes you introduce are breaking the compatibility for current users of the v2 and according to semver, this is not possible.

The current beta version of the client (v3) is living on the `develop` and can integrate breaking changes until the release as stable, which is gonna happen in the weeks to come.

For the record, here's the full review:
 - `deleteRule` exists in v3 in two flavors (with and without `RequestOptions`), because `forwardToReplicas` is supposed to be passed as an URL parameter in that case
 - Fluent API setter changes are perfect, thanks again for that
 - because `QueryBase.addCustomParameter` does not exist anymore in the v3, this change is not applicable anyway.

I've taken the liberty to reproduce your changes (at least the ones being applicable) to another PR based on `develop`: https://github.com/algolia/algoliasearch-client-java-2/pull/573. Feel free to comment and follow on this one.

Thank you again for taking the time to contribute.

PR #596, opened 2019-07-05
adds: GZip compression feature (#591)

### LINKED ISSUE #591: API Feature - Implement GZIP for POST/PUT
The engine is introducing a new feature allowing to optionally compress request
payloads (ie.  POST/PUT body, http or https).

To test it:

Perform a `saveObject` with the content encoded with GZIP
Don't forget to add: `"Content-Encoding: gzip"` as a header.

We will discuss a specification for a more detailed implementation.
Like adding a `enableGZIP` flag in client's configuration.

Ant-hem:
Adds the possibility to compress request for SearchClient.
For the moment only GZIP compression is available.
GZIP compression is enabled by default for SearchClient.

| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #591 
| Need Doc update   | yes

`AccountClient` test is expected to fail because the feature is not enabled on the second test server yet.

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/AnalyticsConfig.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/AnalyticsConfig.java
@@ -1,6 +1,7 @@
 package com.algolia.search;
 
 import com.algolia.search.models.common.CallType;
+import com.algolia.search.models.common.CompressionType;
 import java.util.*;
 import javax.annotation.Nonnull;
 
@@ -16,7 +17,7 @@ public final class AnalyticsConfig extends ConfigBase {
      * @param apiKey The API Key
      */
     public Builder(@Nonnull String applicationID, @Nonnull String apiKey) {
-      super(applicationID, apiKey, createDefaultHosts());
+      super(applicationID, apiKey, createDefaultHosts(), CompressionType.NONE);
     }
 
     @Override
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/HttpRequest.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/HttpRequest.java
@@ -1,11 +1,22 @@
 package com.algolia.search.models;
 
+import com.algolia.search.models.common.CompressionType;
 import java.io.InputStream;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
 
 public class HttpRequest {
+
+  public HttpRequest(
+      HttpMethod method,
+      String methodPath,
+      Map<String, String> headers,
+      int timeout,
+      CompressionType compressionType) {
+    this(method, methodPath, headers, timeout);
+    this.compressionType = compressionType;
+  }
 
   public HttpRequest(
       HttpMethod method, String methodPath, Map<String, String> headers, int timeout) {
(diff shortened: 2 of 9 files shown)


algoliasearch-core/src/main/java/com/algolia/search/AnalyticsConfig.java
```diff
@@ -16,7 +17,7 @@
      * @param apiKey The API Key
      */
     public Builder(@Nonnull String applicationID, @Nonnull String apiKey) {
-      super(applicationID, apiKey, createDefaultHosts());
+      super(applicationID, apiKey, createDefaultHosts(), CompressionType.NONE);
```
aseure, line 20: nit: why is this part of the subclass constructor signature?

algoliasearch-core/src/main/java/com/algolia/search/models/HttpRequest.java
```diff
@@ -69,6 +80,27 @@ public HttpRequest setTimeout(int timeout) {
...
+   * Tells if any compression can be enabled for a request or not. Compression is enabled only for
+   * POST/PUT methods on the Search API (not on Analytics and Insights).
+   */
+  public boolean canCompress() {
+    if (this.compressionType == null || this.method == null) {
+      return false;
+    } else
+      return this.compressionType.equals(CompressionType.GZIP)
+          && (this.method.equals(HttpMethod.POST) || this.method.equals(HttpMethod.PUT));
+  }
```
aseure, line 106: All these nested `if`/`else` could be simplified a bit such as:

```
var isMethodValid = method.equals.(POST) || method.equals(PUT);
var isCompressionEnabled = ! compression.equals(NONE)

return isMethodValid && isCompressionEnabled
```

Also, you mentioned in the comments that should never return true for Analytics and Insights API so you may want to introduce a third boolean here to check that the API kind is the Search API and to add it to the return call.

WDYT?

Ant-hem: > All these nested `if`/`else` could be simplified a bit such as:

Sure let's simplify that 👍 

> Also, you mentioned in the comments that should never return true for Analytics and Insights API so you may want to introduce a third boolean here to check that the API kind is the Search API and to add it to the return call.

Yes, to have a more robust code we should introduce another test. Here is my proposal:

The `HttpRequest` object is taking the `compression` property from the configuration in `HttpTransport`. Let's add an additional control in`configuration.getCompressionType()` to test whether or not it's an Analytic/Insight config instance and return false for both of them. 

WDYT?

aseure: Always all in for more tests 💯

algoliasearch-core/src/main/java/com/algolia/search/AnalyticsConfig.java
aseure: As discussed with @Ant-hem: I missed the point where the `CompressionType.NONE` could actually not be used as default. Sorry for the noise.

algoliasearch-core/src/main/java/com/algolia/search/models/HttpRequest.java
Ant-hem: @aseure I fixed the two comments above with 155ee4a.

`if/else` statement are now simplified. Regarding extra check for the compression type, I found a way to have `setCompressionType()` only avaible on `SearchConfig`. Now there is no way for a user to set a `compressionType` for a config that doesn't support it, such as `Analytics` or `Insights.`

aseure (approved):
I'd say the PR is ready to merge as-is feature wise, hence my approval. Once thing you could consider though is to make the implementation not GZIP-specific. What I mean by that is that we chose to introduce an enum `NONE`/`GZIP` for the compression type instead of a boolean. However, when it comes to serialization, we don't switch among the different possibilities. This is definitely nitpicking as we don't plan to have any other compression format any time soon and it is clearly internal design, hence the approval :) Up to you buddy.

From a functional point of view, super good change! Thank you again for leading this with the engine team @Ant-hem 👍 👍 👍

Ant-hem:
Thanks for the comments @aseure! 🎉 

> What I mean by that is that we chose to introduce an enum `NONE`/`GZIP` for the compression type instead of a boolean.

I did the update with the `enum` for both Java and C# PRs. See for [Java](https://github.com/algolia/algoliasearch-client-java-2/pull/596/files#diff-93445a7b5e8dceb2ebdf8b84f4792b45). Did I miss something else?

PR #598, opened 2019-07-10
fixed(synonym): handle deserialization of lowercase variant of synonym types

### LINKED ISSUE #597: (2.22.0) fix: OneWaySynonym serialization error
The client is expecting `oneWaySynonym` as subtype, however when saving a onewaysynoym from the dashboard the api is sending `onewaysynonym` making the serializer to throw an expection.

aseure:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #597 
| Need Doc update   | no

## Describe your change

    This commit adds new package-private interface to list all lowercase
    variants of the synonym types and use them as JsonSubTypes annotations
    of the AbstractSynonym class.

    Note that we considered using ACCEPT_CASE_INSENSITIVE_PROPERTIES
    MapperFeature from jackson-databind but this needs to be set globally at
    the ObjectMapper's level, which we do not want.

    Synonyms saved through the Algolia dashboard see their types stored as
    lowercase. Jackson not being able to deserialize lowercase enums and
    keys into their upper-case or mixed-case counterparts, retrieving
    synonyms that were saved from the dashboard made implementations crash.

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/inputs/synonym/AbstractSynonym.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/inputs/synonym/AbstractSynonym.java
@@ -11,8 +11,11 @@ import java.io.Serializable;
 )
 @JsonSubTypes({
   @JsonSubTypes.Type(value = AltCorrection1.class, name = SynonymType.ALT_CORRECTION_1),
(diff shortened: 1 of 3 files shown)


Ant-hem (approved):
Thanks for the fix 🎉 The workaround is good.

PR #600, opened 2019-07-10
(2.22.0) Adds: missing query parameters

### LINKED ISSUE #599: (2.22.0) Add missing query parameters
Let's add missing query parameters for v2.X.

Ant-hem:
Fix #599

| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #599
| Need Doc update   | no

### DIFF
--- a/algoliasearch-common/src/main/java/com/algolia/search/objects/QueryBase.java
+++ b/algoliasearch-common/src/main/java/com/algolia/search/objects/QueryBase.java
@@ -32,6 +32,7 @@ import java.util.Map;
 public abstract class QueryBase<T extends QueryBase<?>> implements Serializable {
 
   /* advanced */
+  protected String similarQuery;
   protected Distinct distinct;
   protected Boolean getRankingInfo;
   protected List<String> numericFilters;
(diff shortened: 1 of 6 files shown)


algoliasearch-common/src/main/java/com/algolia/search/objects/QueryBase.java
```diff
@@ -298,6 +303,63 @@
...
+                .stream()
+                .map(
+                    arr ->
+                        "["
+                            + arr.stream()
+                                .map(QueryBase::formatParameters)
+                                .collect(Collectors.joining(","))
+                            + "]")
+                .collect(Collectors.joining(","))
+            + "]";
```
aseure, line 325: I don't know if this style is enforced by `coveo` but it is particularly difficult to read. May we try to change that?

aseure (commented):
Great changes, thank you Antoine!

Only comment but maybe too late: we could have split this in multiple commits (one of each type added/changed and one for the tests) and some for the tests. This way, once we generate the ChangeLog entry, we could directly let users see what has changed (especially as we are introducing breaking changes). WDYT?

algoliasearch-common/src/main/java/com/algolia/search/objects/QueryBase.java
Ant-hem: It's coveo styling convention and I couldn't find a way to change it.

aseure (approved):

PR #609, opened 2019-07-24
fix: missing catch clauses in the httpRequester

Ant-hem:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

The requester was not catching correctly all exceptions, making the
retry strategy to fail instead of retrying for network related issues.

Apache HTTP client related materials:

Documentation about exception thrown by Apache HTTP client:
https://hc.apache.org/httpcomponents-core-4.4.x/tutorial/html/blocking-io.html#d5e375

(Legacy doc but some exceptions are still up-to-date)
https://hc.apache.org/httpclient-3.x/exception-handling.html

### DIFF
--- a/algoliasearch-apache/src/main/java/com/algolia/search/ApacheHttpRequester.java
+++ b/algoliasearch-apache/src/main/java/com/algolia/search/ApacheHttpRequester.java
@@ -12,15 +12,14 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import javax.annotation.Nonnull;
-import org.apache.http.Header;
-import org.apache.http.HeaderElement;
-import org.apache.http.HttpEntity;
+import org.apache.http.*;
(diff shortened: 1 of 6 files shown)


Ant-hem:
@aseure & @BenoitPerrot : I couldn't find a "source of truth" for all the potentials exceptions thrown by the `HttpAsyncClient` of Apache. I think we are handling all cases here, but I'd like to have a double check. Thanks!

aseure (approved):
Cannot tell you about extra `Exception` to look for. Good catch and thank you for taking care of those already!

BenoitPerrot (approved):
For Another Review: could we leverage the info `isNetworkError` for emitting a better diagnostic when such thing occur?

Ant-hem:
> For Another Review: could we leverage the info `isNetworkError` for emitting a better diagnostic when such thing occur?

Definitely, I had a chat with Adam about improving our errors messages when all hosts are unreachable, I am going to open an issue in our spec repository.

PR #621, opened 2019-08-13
feat(is secured api key expired)

### LINKED ISSUE #612: New method: GetSecuredApiKeyRemainingValidity
## Summary

When a secured API keys expire, search breaks. The current workaround is more complex
than it needs to be. 

The options for developers using Algolia require, at every place where a search
operation is performed, catching an operation error, identify it as a
"ValidUntil" error, then regenerate the API key, and perform the search
operation again.

The idea is to define a method which tells if a `SecuredAPIKey`
is expired or not without making a single call to Algolia's API.

## References

- Specs: https://github.com/algolia/algoliasearch-client-specs-internal/pull/25

Ant-hem:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #612  
| Need Doc update   | yes

## Describe your change
See #612

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/SearchClientAPIKeys.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/SearchClientAPIKeys.java
@@ -10,9 +10,13 @@ import com.algolia.search.models.apikeys.*;
 import com.algolia.search.models.common.CallType;
 import com.algolia.search.util.AlgoliaUtils;
 import com.algolia.search.util.HmacShaUtils;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Base64;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
+import java.util.regex.*;
 import javax.annotation.Nonnull;
 
 public interface SearchClientAPIKeys extends SearchClientBase {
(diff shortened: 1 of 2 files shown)


algoliasearch-core/src/main/java/com/algolia/search/SearchClientAPIKeys.java
```diff
@@ -422,4 +426,35 @@ default String generateSecuredAPIKey(
...
+  default Duration getSecuredApiKeyRemainingValidity(@Nonnull String securedAPIKey) {
+
+    if (AlgoliaUtils.isNullOrEmptyWhiteSpace(securedAPIKey)) {
+      throw new AlgoliaRuntimeException("securedAPIKey must not be empty, null or whitespaces");
+    }
+
+    byte[] decodedBytes = Base64.getDecoder().decode(securedAPIKey);
+    String decodedString = new String(decodedBytes);
+
+    Pattern pattern = Pattern.compile("validUntil=\\d{10}");
```
aseure: Why is there a 10 digits limit? Can't we use `\\d+` instead?

Ant-hem: I choose `\\d{10}` because the UNIX timestamp returned by the engine will have 10 digits length! Do you think it's too restrictive and not future proof?

aseure (approved):

PR #650, opened 2019-10-25
feat(query_rules): adds missing enabled parameter

Ant-hem:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

adds missing `enabled` parameter in `RuleQuery`

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/rules/RuleQuery.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/rules/RuleQuery.java
@@ -16,7 +16,7 @@ public class RuleQuery implements Serializable {
   private String query;
   private String anchoring = null;
   private String context = null;
-
+  private Boolean enabled;
   private Long page = null;
   private Long hitsPerPage = null;
 
@@ -84,6 +84,15 @@ public class RuleQuery implements Serializable {
   /** Set the number of hits per page. Defaults to 10. */
   public RuleQuery setHitsPerPage(Integer hitsPerPage) {
     return this.setHitsPerPage(hitsPerPage.longValue());
+  }
+
+  public Boolean getEnabled() {
+    return enabled;
+  }
+
+  public RuleQuery setEnabled(Boolean enabled) {
+    this.enabled = enabled;
+    return this;
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


algoliasearch-core/src/main/java/com/algolia/search/models/rules/RuleQuery.java
```diff
@@ -16,7 +16,7 @@
   private String query;
   private String anchoring = null;
   private String context = null;
-
+  private Boolean enabled;
```
aseure, line 19: Shouldn't we add a default value of `true`?

Ant-hem: According to the doc `enabled` is optional: https://www.algolia.com/doc/api-reference/api-methods/search-rules/#method-param-enabled

So a `nullable` Boolean does the trick :)

That's also what we have in C#:  

https://github.com/algolia/algoliasearch-client-csharp/blob/a50b5330e747aa3ba8f735d9296fe542d6f0d6c9/src/Algolia.Search/Models/Rules/RuleQuery.cs#L69

aseure: Ok got it 👍 Then should we make it `private Boolean enabled = null;`?

Ant-hem: `private Boolean enabled;` is redundant with `private Boolean enabled = null;`
Others parameters like `private String context = null;` are like this because I C/C from the `v2`. I could update them to be consistent what do you think?

aseure: Ok, thank you for the context. Then yes, we should probably remove the `= null` in that case. This is what confused me in the first place, and it will probably bite us one more time in the future.

aseure (approved):
I've approved this PR so that you're not blocked on this one. All is good to me.

algoliasearch-core/src/main/java/com/algolia/search/models/rules/RuleQuery.java
Ant-hem: Fixed with 74d33b2. Let me know if it's good for you 👍

PR #664, opened 2019-11-25
patch: serialization/deserialization of ConsequenceQuery

### LINKED ISSUE #662: investigate(v3/query_rules) replace query
- Algolia Client Version: 3

### Description

Make sure `params.query` is correctly serialized and deserialized.

Json Example:
```json
{"params":{"query":"toto"}}
```

Ant-hem:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #662 
| Need Doc update   | no

## Problem
The custom serializer was not handling all the cases for the payload below.

Example:

```json
// query string
"query": "some query string"

// remove attribute (deprecated)
"query": {"remove": ["query1", "query2"]}

// edits attribute
"query": {
   "edits": [
     { "type": "remove", "delete": "old" },
     { "type": "replace", "delete": "new", "insert": "newer" }
   ]
}}
```

## Fix

I added a `string` property in `ConsequenceQuery` and modified the serializer to allow `ConsequenceQuery` to deserialize and serializer Query as string. This required to overload `setQuery` and `getQuery` from `SearchParameters`

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/rules/ConsequenceQuery.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/rules/ConsequenceQuery.java
@@ -1,14 +1,19 @@
 package com.algolia.search.models.rules;
 
 import com.algolia.search.Defaults;
+import com.algolia.search.util.AlgoliaUtils;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.ObjectCodec;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.*;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -19,9 +24,10 @@ import java.util.stream.Collectors;
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonDeserialize(using = ConsequenceQueryDeserializer.class)
+@JsonSerialize(using = ConsequenceQuerySerializer.class)
 public class ConsequenceQuery implements Serializable {
 
-  private List<Edit> edits;
+  public ConsequenceQuery() {}
 
   public List<Edit> getEdits() {
     return edits;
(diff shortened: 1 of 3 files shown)


nunomaduro:
@Ant-hem Thanks for this. There any breaking change concerning what we had before?  Or his just adding a new feature for a particular situation that wasn't possible before?

Ant-hem:
> @Ant-hem Thanks for this. There any breaking change concerning what we had before? Or his just adding a new feature for a particular situation that wasn't possible before?

It's a backwards compatible bug fix. To give more context:  before the patch, using `setQuery` (in `Rule.ConsequenceQuery`) would have produced an empty payload. Now it's writing the correct one. So I had to modify both serializer and deserializer.

nunomaduro:
Great! Let's wait for @aseure to review the code.

algoliasearch-core/src/main/java/com/algolia/search/models/rules/ConsequenceQuery.java
```diff
@@ -31,29 +37,118 @@ public ConsequenceQuery setEdits(List<Edit> edits) {
...
+class ConsequenceQuerySerializer extends JsonSerializer<ConsequenceQuery> {
+  @Override
+  public void serialize(ConsequenceQuery value, JsonGenerator gen, SerializerProvider serializers)
+      throws IOException {
+    /*
+     * Consequence query edits will override regular "query" - both can't be set at the same time
+     * https://www.algolia.com/doc/api-reference/api-methods/save-rule/#method-param-query
+     * */
+    if (!AlgoliaUtils.isNullOrEmptyWhiteSpace(value.getQueryString())) {
+      gen.writeString(value.getQueryString());
```
aseure, line 84: You may want to migrate `value.getQueryString()` in its own temporary: it's probably negligible in term of performances but for readability, it would help at least.

aseure (approved):

algoliasearch-core/src/main/java/com/algolia/search/models/rules/ConsequenceQuery.java
Ant-hem: > You may want to migrate `value.getQueryString()` in its own temporary: it's probably negligible in term of performances but for readability, it would help at least.

Could you elaborate? i don't get it.

aseure: I was suggesting something like this:

```java
String query = value.getQueryString();

 if (!AlgoliaUtils.isNullOrEmptyWhiteSpace(query)) {
      gen.writeString(query);
} else {
     // ...
}
```

aseure (approved):
Approving the change regarding the override of the `ConsequenceQuery` when calling `.setQuery()`.

PR #665, opened 2019-12-06
feat: adds JDK11 java-net requester

Ant-hem:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

The PR adds a new `algoliasearch-java-net` module which embeds the JDK11 http requester. This allow our JDK11 users to drop the dependency on Apache and to profit from HTTP2.

Adding a JDK11 module with other JDK8 modules require to update the `java-doc` plugin and to add `<source>1.8</source>` in its configuration, otherwise the build will fail. (Known issue on open JDK)

The CI is also updated: for JDK8; Travis will only run the test for the Apache module; for JDK11 it will test both `java-net` and `apache` modules.

## Review guidelines

Please review thoroughly the `java-net` requester
Please review thoroughly all pom.xml

### DIFF
--- a/.run-travis.sh
+++ b/.run-travis.sh
@@ -18,7 +18,7 @@ if [[ "${JAVA_VERSION}" = "11" ]]; then
     eval $(./algolia-keys export) && mvn clean compile && mvn clean test;
     exit $?
   else
-    mvn clean compile & mvn clean test;
+    mvn clean compile && mvn clean test;
     exit $?
   fi
 fi
--- a/.run-travis.sh
+++ b/.run-travis.sh
@@ -4,20 +4,21 @@ set -e
 
 if [[ "${JAVA_VERSION}" = "8" ]]; then
   if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [[ ! "$TRAVIS_PULL_REQUEST_SLUG" =~ ^algolia\/ ]]; then
-    eval $(./algolia-keys export) && mvn clean test;
+    # test only JDK8 comptaible requester
+    eval $(./algolia-keys export) && mvn clean compile -pl algoliasearch-apache,algoliasearch-core &&  mvn clean test -pl algoliasearch-apache,algoliasearch-core;
     exit $?
   else
-    mvn clean test;
+    mvn clean compile -pl algoliasearch-apache,algoliasearch-core && mvn clean test -pl algoliasearch-apache,algoliasearch-core;
     exit $?
   fi
 fi
 
 if [[ "${JAVA_VERSION}" = "11" ]]; then
   if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [[ ! "$TRAVIS_PULL_REQUEST_SLUG" =~ ^algolia\/ ]]; then
-    eval $(./algolia-keys export) && mvn clean test;
+    eval $(./algolia-keys export) && mvn clean compile && mvn clean test;
     exit $?
   else
-    mvn clean test;
+    mvn clean compile & mvn clean test;
     exit $?
   fi
 fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,11 @@ matrix:
     env: JAVA_VERSION=8
   - jdk: openjdk11
     env: JAVA_VERSION=11
-
 cache:
   directories:
   - $HOME/.m2
+
+install: skip # skipping default travis setup to support both JDK8 and JDK11 build. Default build would fail.
 
 #check coverage
 before_script:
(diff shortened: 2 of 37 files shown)


.travis.yml
```diff
@@ -6,11 +6,12 @@ matrix:
     env: JAVA_VERSION=8
   - jdk: openjdk11
     env: JAVA_VERSION=11
-
 cache:
   directories:
   - $HOME/.m2
 
+install: skip # skipping default travis setup to support both JDK8 and JDK11 build. Default build would fail.
```
aseure, line 13: Thanks for the comment 👍

.run-travis.sh
```diff
@@ -4,20 +4,21 @@ set -e
...
 fi
 
 if [[ "${JAVA_VERSION}" = "11" ]]; then
   if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [[ ! "$TRAVIS_PULL_REQUEST_SLUG" =~ ^algolia\/ ]]; then
-    eval $(./algolia-keys export) && mvn clean test;
+    eval $(./algolia-keys export) && mvn clean compile && mvn clean test;
     exit $?
   else
-    mvn clean test;
+    mvn clean compile & mvn clean test;
```
aseure: Running `programA & programB` makes `programA` starts in the background and immediately starts `programB` as well. Are you sure this is what you want here?

Don't you prefer `programA && programB` (runs `programB` iff. `programA` terminates successfully) or `programA ; programB` (runs `programB` after `programA`, no matter what)?

You may also simply want to run `mvn clean compile test` (if it works, I'm not sure of the exact behavior).

algoliasearch-core/src/test/java/com/algolia/search/integration/TestHelpers.java
```diff
@@ -5,6 +5,15 @@
...
 
 public class TestHelpers {
+
+  public static String ALGOLIA_APPLICATION_ID_1 = System.getenv("ALGOLIA_APPLICATION_ID_1");
+  public static String ALGOLIA_API_KEY_1 = System.getenv("ALGOLIA_ADMIN_KEY_1");
+  public static String ALGOLIA_SEARCH_KEY_1 = System.getenv("ALGOLIA_SEARCH_KEY_1");
+  public static String ALGOLIA_APPLICATION_ID_2 = System.getenv("ALGOLIA_APPLICATION_ID_2");
+  public static String ALGOLIA_API_KEY_2 = System.getenv("ALGOLIA_ADMIN_KEY_2");
+  public static String ALGOLIA_APPLICATION_ID_MCM = System.getenv("ALGOLIA_APPLICATION_ID_MCM");
+  public static String ALGOLIA_ADMIN_KEY_MCM = System.getenv("ALGOLIA_ADMIN_KEY_MCM");
```
aseure, line 15: I'd suggest renaming all the variables with the same name of the environment variable they are bound to. This would lead to less confusing and back-and-forth when debugging from my own experience. (`Shift+F6` to rename everywhere in IntelliJ ;) )

algoliasearch-core/src/test/java/com/algolia/search/integration/TestHelpers.java
```diff
@@ -19,4 +28,34 @@ public static String getMcmUserId() {
     return String.format(
         "java-%s-%s", DateTimeFormatter.ofPattern("yyyy-MM-dd-HH-mm-ss").format(utc), userName);
   }
+
+  public static void checkEnvironmentVariable() throws Exception {
+    if (ALGOLIA_APPLICATION_ID_1 == null || ALGOLIA_APPLICATION_ID_1.isEmpty()) {
+      throw new Exception("ALGOLIA_APPLICATION_ID is not defined or empty");
```
aseure: Following the previous comment about renaming all variables with the same name as their respective environment variable, I'd suggest to also change the name accordingly in all those thrown exceptions to prevent any confusion.

aseure (changes requested):
Great PR and thanks for the pair-programming session while reviewing it. I've left some small suggestions and requested changes but otherwise, L(super)GTM 👍 🎉

.run-travis.sh
Ant-hem: Good catch :D It's a typo.

algoliasearch-core/src/test/java/com/algolia/search/integration/TestHelpers.java
Ant-hem: You're right, let's stay consistent. I'll do it.

aseure (approved):

PR #673, opened 2020-03-10
fix(account): use correct AlgoliaRuntimeException message when copying an index to an already existing destination

aseure:


Ant-hem (approved):
Thanks!

PR #675, opened 2020-03-11
fix(synonym): correctly serialize SynonymQuery's type attribute

### LINKED ISSUE #674: Ensure correct serialization of searchSynonyms' type query parameter
When performing a `index.searchSynonyms()` call, a [`type` query
parameter](https://www.algolia.com/doc/rest-api/search/#method-param-type) can
be passed to restrict the type(s) of synonyms one wants to search for.

@maxiloc reported us that the Java implementation is accepting an array of
strings for this parameter. However, the engine only accepts a comma-separated
string for this. We then need to double check the implementation of the
serialization of this parameter and provide a fix if needed.

aseure:
Fix #674

### DIFF
--- a/algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
+++ b/algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
@@ -1,5 +1,7 @@
 package com.algolia.search;
 
+import static com.algolia.search.models.synonyms.SynonymType.ALT_CORRECTION_1;
+import static com.algolia.search.models.synonyms.SynonymType.ONE_WAY_SYNONYM;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.algolia.search.models.indexing.AroundPrecision;
@@ -7,6 +9,7 @@ import com.algolia.search.models.indexing.AroundRadius;
 import com.algolia.search.models.indexing.Query;
 import com.algolia.search.models.rules.*;
 import com.algolia.search.models.settings.*;
+import com.algolia.search.models.synonyms.SynonymQuery;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import java.io.IOException;
 import java.util.*;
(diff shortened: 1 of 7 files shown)


algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
```diff
@@ -806,4 +809,12 @@ private IndexSettings serializeDeserialize(IndexSettings obj) throws IOException
             serialized,
             Defaults.getObjectMapper().getTypeFactory().constructType(IndexSettings.class));
   }
+
+  @Test
+  void searchSynonymsQueryWithType() throws JsonProcessingException {
+    SynonymQuery query =
+        new SynonymQuery("query").setType(Arrays.asList(ONE_WAY_SYNONYM, ALT_CORRECTION_1));
+    String serialized = Defaults.getObjectMapper().writeValueAsString(query);
+    assertThat(serialized).isEqualTo("{\"query\":\"q\",\"type\":\"oneWaySynonym,altCorrection1\"}");
```
Ant-hem, line 818: Does the test pass? IIRC the default separator for `FlatListSerializer` is a `;`. We could parametrize it.

https://github.com/algolia/algoliasearch-client-java-2/blob/febbaeafd7fbdee4affc9856a4601947a1bd1443/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FlatListSerializer.java#L19-L24

aseure: Indeed it does not. My command `mvn -DfailIfNoTests=false -Dtest=com.algolia.search.searchSynonymsQueryWithType test` to run a single test is not correct (hence the test was not run and everything was green).

aseure: For the record, here's the correct maven command (using `class#method`):
```
mvn -DfailIfNoTests=false -Dtest='JacksonParserTest#searchSynonymsQueryWithType' test
```

aseure:
Waiting for some answers from the search team regarding the exact types of the currently `FlatListSerializer`-serialized types to implement the right fix (and potential type corrections).

aseure:
@Ant-hem Finished it following the confirmation we got from the search team regarding all the types.

Ant-hem (approved):
LGTM 👍 Thanks a lot.

PR #690, opened 2020-04-16
Make Apache HTTP client more configurable and improve default performances

### LINKED ISSUE #678: Let users configure the Apache client the via the ApacheHttpRequester() constructor
@Ant-hem and I recently investigated a performance bottleneck for a customer which led us to consider exposing the `HttpAsyncClientBuilder` use to set the [`ApacheHttpRequester.asyncHttpClient`](https://github.com/algolia/algoliasearch-client-java-2/blob/c87e6f2c654677a3de707ab13dd1f470e2c4588f/algoliasearch-apache/src/main/java/com/algolia/search/ApacheHttpRequester.java#L36).

A first idea would be to deprecate [the use of `useSystemProxy`](https://github.com/algolia/algoliasearch-client-java-2/blob/c87e6f2c654677a3de707ab13dd1f470e2c4588f/algoliasearch-apache/src/main/java/com/algolia/search/ApacheHttpRequester.java#L50-L53).

Then, we'd like to create [a new constructor for the ApacheHttpRequester](https://github.com/algolia/algoliasearch-client-java-2/blob/c87e6f2c654677a3de707ab13dd1f470e2c4588f/algoliasearch-apache/src/main/java/com/algolia/search/ApacheHttpRequester.java#L40-L56) with the following signature:

```java
ApacheHttpRequester(@Nonnull ConfigBase config, @Nonnull HttpAsyncClientBuilder builder) {
    // ...
}
```

This way, users could completely override the default Apache HTTP client configuration, such as `builder.setMaxConnPerRoute()` and `builder.setMaxConnPerRoute()`. Side-note: we may also want to increased the default values for at least `maxConnPerRoute` from 2 (default value) to something that would make more sense for real-life situations, as explained in the [Apache HTTP client documentation](https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html):

Screenshot 2020-03-13 at 11 48 39

For the record, our benchmarks showed that changing the `maxConnPerRoute` from 2 to 100, the number of req/s jumped from ~100 to ~1000, which makes me think we should also consider changing the default value for our default client as well.

aseure:
Fix #678


algoliasearch-core/src/main/java/com/algolia/search/ConfigBase.java
```diff
@@ -82,6 +82,7 @@ public Builder(
     public abstract T getThis();
 
     /** Makes the underlying HTTP Client to use JVM/System settings for the proxy */
+    @Deprecated
```
Ant-hem: I would add a note in the deprecation attribute telling to use the `builder` instead.

algoliasearch-apache/src/main/java/com/algolia/search/ApacheHttpRequester.java
```diff
@@ -31,14 +32,13 @@
...
-final class ApacheHttpRequester implements HttpRequester {
+public final class ApacheHttpRequester implements HttpRequester {
 
   private final CloseableHttpAsyncClient asyncHttpClient;
   private final RequestConfig requestConfig;
   private final ConfigBase config;
 
-  ApacheHttpRequester(@Nonnull ConfigBase config) {
-
+  public ApacheHttpRequester(@Nonnull ConfigBase config) {
```
Ant-hem, line 40: nit: We could call `ApacheHttpRequester(@Nonnull ConfigBase config, @Nonnull HttpAsyncClientBuilder builder)` with the builder created in place in `ApacheHttpRequester(@Nonnull ConfigBase config)` to avoid duplicated code in the ctors.

aseure: I wanted to do it at first but then abandoned because I needed to pass the big ternary in a `this( ... );` statement which must be the first immediate statement in the constructor to be able to chain them. But after retrying it, it seems kinda clean anyway so I've did it in the end.

algoliasearch-core/src/main/java/com/algolia/search/ConfigBase.java
aseure: `@Deprecated` annotation doesn't support optional attributes before Java 9.

I'll put a comment on the same line instead!

Ant-hem: I've done this like this before:

https://github.com/algolia/algoliasearch-client-java-2/blob/6289d61dd96bad7dd3c298dd8b4c24e1ab1bc0ce/algoliasearch-core/src/main/java/com/algolia/search/SearchClientPersonalization.java#L19-L27

aseure: OH, even better, I didn't remember of the javadoc way of doing it. Thanks! Changing it right now.

Ant-hem (approved):
Awesome, thanks! Let's go.

Ant-hem (approved):
Let's go! Thanks for migrating to Circle.

PR #696, opened 2020-06-03
Feat/expose built in ops partial update

aseure:


algoliasearch-core/src/main/java/com/algolia/search/models/indexing/PartialUpdateOperation.java
```diff
@@ -0,0 +1,53 @@
+package com.algolia.search.models.indexing;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class PartialUpdateOperation<T extends Object> {
```
aallam: No need to explicistly extend `Object`, we can replace `T extends Object` by `T`

algoliasearch-core/src/main/java/com/algolia/search/models/indexing/PartialUpdateOperation.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public T getValue() {
+    return value;
+  }
+
+  public void setValue(T value) {
+    this.value = value;
+  }
+
+  public static PartialUpdateOperation<Integer> Increment(Integer value) {
```
aallam: Maybe we should avoid capital letters for method names?
(same applies to other methods below)

We can either use lowercase letters for the methods, or replace them with inner static classes:
```java
public static class Increment extends PartialUpdateOperation<Integer> {
    public Increment(Integer value) {
      super(PartialUpdateOperationType.INCREMENT, value);
    }
}
```

aseure: Right, thanks a lot! I'm doing Go, Scala, C# and Java at the same time today so my casing game is a bit weird today ^^

I'll go with keeping the static functions if that's OK with you, I'm not a big fan of introducing inheritance for such a niche use-case.

aallam (approved):
Looks perfect to me 👍

aseure:
@Ant-hem @Aallam Just fixed the build with a third commit (see commit message for the details). Everything's still good to you?

aallam (approved):
Yup all good 👍

Ant-hem (approved):

PR #697, opened 2020-06-05
feat(build): apache-uber.jar: make it loadable as an OSGi bundle

BenoitPerrot:
By adding OSGi metadata to its manifest.

The generated (uber) jar can then be loaded in an OSGi platform. For example, it can be `installed` in Apache Felix. Note that to be "absolutely safe", we should include an (integration) test installing the generated jar in Felix along with an OSGi bundle leveraging the classes.

### DIFF
--- a/algoliasearch-apache-uber/pom.xml
+++ b/algoliasearch-apache-uber/pom.xml
@@ -18,6 +18,27 @@
     <build>
         <plugins>
             <plugin>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                    <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+                    </archive>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <executions>
+                    <execution>
+                    <id>bundle-manifest</id>
+                    <phase>process-classes</phase>
+                    <goals>
+                        <goal>manifest</goal>
+                    </goals>
+                    </execution>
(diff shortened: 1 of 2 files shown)


BenoitPerrot:
Looking into testing the setup.

After comparison between junit4osgi and pax-exam, the latter seems more promising. See also https://github.com/ops4j/org.ops4j.pax.exam2

algoliasearch-apache-uber/pom.xml
```diff
@@ -17,6 +17,27 @@
     </dependencies>
     <build>
         <plugins>
+            <plugin>
+                <artifactId>maven-jar-plugin</artifactId>
```
aseure, line 21: Could you please add a `<version>2.4</version>` just after the `<artifactId>` declaration. Otherwise, a warning is generated at each Maven run.

I know there are more recent versions of `maven-jar-plugin` but since `2.4` is the one we use in the top `pom.xml` file, I suggest to keep the same version for this PR and open a new PR whenever we will want to upgrade it.

aseure (changes requested):

aseure:
As a personal reminder, here's the diff we get between the two `MANIFEST.MF`:

Screenshot 2020-06-08 at 11 13 44

aseure:
Following the article you found @BenoitPerrot we do not seem to generate the `Bundle-ClassPath` nor the `Export-Package` entries. I'll continue to learn about OSGi bundles and see, as you recommended, how we can test it properly.

Thanks for looking up at it first @BenoitPerrot! Really appreciate it.

BenoitPerrot:
I'm not very proud of the status of what I shared, but I'm happy if the hints were somehow useful!

aseure:
Work continued on #701

PR #698, opened 2020-06-08
chore(compiler): explitly specify source and target versions

aallam:
| Q                 | A
| Bug fix?          |  no
| New feature?      | no   
| BC breaks?        | no     

## Describe your change

Explicitly specify source and target version of the `algoliasearch-java-net` module

Remove warnings about the use of Java APIs available only since version 11

### DIFF
--- a/algoliasearch-java-net/pom.xml
+++ b/algoliasearch-java-net/pom.xml
@@ -9,19 +9,11 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
+    <properties>
+        <maven.compiler.release>11</maven.compiler.release>
+    </properties>
+
     <artifactId>algoliasearch-java-net</artifactId>
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <source>11</source>
-                    <target>11</target>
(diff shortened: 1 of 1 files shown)


Ant-hem (approved):
Sweet, thanks!

aseure (approved):

algoliasearch-java-net/pom.xml
```diff
@@ -10,6 +10,18 @@
...
 
     <artifactId>algoliasearch-java-net</artifactId>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>11</source>
+                    <target>11</target>
```
aseure: Since Java 9, the `-release` option now supersedes both `-source` and `-target` and sets other flags correctly.

You can replace both `<source>` and `<target>` with `<release>11</release>`.

aseure (changes requested):

algoliasearch-java-net/pom.xml
aallam: Didn't know about it! thanks!
The `maven-compiler-plugin` seem to have this feature only starting from version 3.6 (we use 3.5.1), luckily, we can use plugin properties instead :D

aseure (approved):

PR #700, opened 2020-06-08
feat(settings): accept enablePersonalization boolean as a valid setti…

### LINKED ISSUE #693: Accept `enablePersonalization` boolean as a valid setting parameter
All API clients currently should accept the `enablePersonalization` boolean
as a valid search parameter, which defaults to `false` when missing. However,
@rayrutjes recently reported that this parameter is also part of the settings
of an index.

The needed change is to expose the `enablePersonalization` boolean as a
valid setting parameter, which defaults to `false` when missing. It still
needs to be accepted as a search parameter as well.

aallam:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #693
| Need Doc update   | yes

## Describe your change

Accept `enablePersonalization` boolean as a valid setting parameter

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
@@ -29,6 +29,7 @@ public class IndexSettings implements Serializable {
   private List<String> queryLanguages;
   private List<String> indexLanguages;
   private Map<String, Map<String, String>> customNormalization;
+  private Boolean enablePersonalization;
 
   private List<String> searchableAttributes;
   private List<String> attributesForFaceting;
@@ -668,6 +669,20 @@ public class IndexSettings implements Serializable {
   public IndexSettings setCustomSettings(Map<String, Object> customSettings) {
     this.customSettings = customSettings;
     return this;
+  }
+
+  public Boolean getEnablePersonalization() {
+    return enablePersonalization != null && enablePersonalization;
+  }
+
+  public IndexSettings setEnablePersonalization(Boolean enablePersonalization) {
+    this.enablePersonalization = enablePersonalization;
+    return this;
+  }
+
+  @JsonProperty("enablePersonalization")
+  private Boolean internalEnablePersonalization() {
+    return enablePersonalization;
   }
 
   @Override
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
@@ -672,17 +672,12 @@ public class IndexSettings implements Serializable {
   }
 
   public Boolean getEnablePersonalization() {
-    return enablePersonalization != null && enablePersonalization;
+    return enablePersonalization;
   }
 
   public IndexSettings setEnablePersonalization(Boolean enablePersonalization) {
     this.enablePersonalization = enablePersonalization;
     return this;
-  }
-
-  @JsonProperty("enablePersonalization")
-  private Boolean internalEnablePersonalization() {
-    return enablePersonalization;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
```diff
@@ -670,6 +672,20 @@ public IndexSettings setCustomSettings(Map<String, Object> customSettings) {
...
+    return enablePersonalization != null && enablePersonalization;
+  }
+
+  public IndexSettings setEnablePersonalization(Boolean enablePersonalization) {
+    this.enablePersonalization = enablePersonalization;
+    return this;
+  }
+
+  @JsonProperty("enablePersonalization")
+  private Boolean internalEnablePersonalization() {
```
aallam: I wonder if this is a good approach? It's just a way to not serialize/deserialize then the property when not set, however, still get the default value `false` from the getter when the property is not set.

What do you think? we can also always serialize the property, but not sure if it is the wanted behavior!

algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
```diff
@@ -670,6 +671,20 @@ public IndexSettings setCustomSettings(Map<String, Object> customSettings) {
...
+  public Boolean getEnablePersonalization() {
+    return enablePersonalization != null && enablePersonalization;
+  }
+
+  public IndexSettings setEnablePersonalization(Boolean enablePersonalization) {
+    this.enablePersonalization = enablePersonalization;
+    return this;
+  }
+
+  @JsonProperty("enablePersonalization")
```
Ant-hem: You don't need `@JsonProperty` here because Jackson will (by default) serialize/deserialize the property with its name, here defined line 32.

algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
Ant-hem: I don't think we need that because `enablePersonalization` is a `Boolean`, an Object and thus nullabe.
At the top of the class we have set `JSON.Include_non_null` so all null values won't be serialized. A simple getter/setter will work for that case, like `advancedSyntax` for example:

https://github.com/algolia/algoliasearch-client-java-2/blob/147aea5f277db9016afe48516fae8fb8ec5ef493/algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java#L357-L364

Does that answer your question?

algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
aallam: The problem is that the property will always get serialized either because it has been set or the default value, the following test, for example, will start to fail: https://github.com/algolia/algoliasearch-client-java-2/blob/92e22dadba7e1b2e589f20688003a2e34ffd54e0/algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java#L29

please check https://github.com/algolia/algoliasearch-client-java-2/pull/700#discussion_r436768692 and tell me what you think?

algoliasearch-core/src/main/java/com/algolia/search/models/settings/IndexSettings.java
aallam: > The needed change is to expose the enablePersonalization boolean as a
valid setting parameter, which defaults to **false** when missing.

It is okay then to not have a default value ?

Ant-hem: Yes! The field won't be serialized and and then engine won't set it internally.

aallam: Nice! even better then :D
Thank you!

Ant-hem (approved):
Sweet, thanks again for your contribution!

aseure (approved):
LGTM 👍

aseure:
@Aallam Could you fix the latest conflicts, and then feel free to `Rebase and Merge`, everything LGTM 👍

PR #703, opened 2020-06-10
fix(search): add support of quoted booleans in IgnorePlurals

### LINKED ISSUE #695: Ensure consequence.params.ignorePlurals can be deserialized from string
It was discovered that the Algolia Search API is sometimes returning the
`ignorePlurals` field from a Query Rule's consequence params field as a
string, such as:

```json
"consequence": {
    "params:" {
        "ignorePlurals": "false"
    }
}
```

Could you please ensure that the deserialization of query rules coming from
`getRule`, `getRules`, `browseRules` are correctly decoding those strings
without failing.

Also, like we did for other string-based booleans, it would be convenient to
convert the string to an actual boolean when the string is equal to "true" or
"false".

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #695
| Need Doc update   | no

## Describe your change

Add support of quoted booleans in `IgnorePlurals`


aseure (approved):
Thanks @Aallam 👍 

I did not remember about Jackson's `BooleanDeserializer`!

PR #704, opened 2020-06-10
fix(explain): add support of `type` property as a List in `Alternative`

### LINKED ISSUE #656: bug: "type" in "Alternatives" should be string[] instead of string
- Algolia Client Version: 3.X
- Language Version:

### Description

The `Explain.Alternative` object has a type parameter that is now a `List<String>` instead of a `String`. The fix requires a custom Deserializer to avoid breaking changes.

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #656
| Need Doc update   | no

## Describe your change

Add a custum deserialization for the `type` property in the Explain.Alternative object.


aseure (approved):
LGTM

PR #705, opened 2020-06-18
test(recommendation): update CTS test with setPersonalizationStrategy call

### LINKED ISSUE #671: tests(cts): adds personalization tests
To solve this issue you need to implement the following cts test: https://github.com/algolia/algoliasearch-client-specs-internal/pull/93/files#diff-5ef12c50215bec6c93882f66c93ba8abR1113

Example in JavaScript : https://github.com/algolia/algoliasearch-client-javascript/pull/1054.
Example in Python : https://github.com/algolia/algoliasearch-client-javascript/pull/1054

aseure:
Close #671

### DIFF
--- a/algoliasearch-core/src/test/java/com/algolia/search/integration/recommendation/RecommendationTest.java
+++ b/algoliasearch-core/src/test/java/com/algolia/search/integration/recommendation/RecommendationTest.java
@@ -1,6 +1,6 @@
 package com.algolia.search.integration.recommendation;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
 import com.algolia.search.Defaults;
 import com.algolia.search.RecommendationClient;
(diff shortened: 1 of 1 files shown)


algoliasearch-core/src/test/java/com/algolia/search/integration/recommendation/RecommendationTest.java
```diff
@@ -25,6 +22,16 @@ protected RecommendationTest(RecommendationClient recommendationClient) {
 
   @Test
   void testRecommendationClient() {
+    SetStrategyRequest request =
```
nunomaduro, line 27: I think this test can be a little bit better.

Maybe you can do like this: https://github.com/algolia/algoliasearch-client-javascript/blob/3991e821b46c860e1badda6dbfecadd3d73701d3/packages/client-recommendation/src/__tests__/features/personalization-strategy.test.ts#L48.

aseure: Got it! Will change it so that the payload is also checked for equality upon reception.

nunomaduro (approved):

PR #709, opened 2020-06-26
fix(build): make the algoliasearch-apache-uber JAR compliant with OSGi

aseure:


aseure:
After a few tries, this version is confirmed to work on our AEM environment as well as the customer's environment.

I would have loved to find a cleaner way to expose the relocated `nio.client` but this seems really difficult since
 1. `maven-bundle-plugin` runs before the `maven-shade-plugin` which relocates several `org.apache.http` packages
 2. @BenoitPerrot' suggestion to change the order of the plugin by changing their `<phase>` lead to another suite of issues, I'm not saying this is impossible, but at this point the current solution is the more efficient in term on invested time.

algoliasearch-apache-uber/pom.xml
```diff
@@ -44,8 +44,13 @@
                     <instructions>
                         <Bundle-SymbolicName>com.algolia.algoliasearch-apache-uber</Bundle-SymbolicName>
                         <Bundle-Name>algoliasearch-apache-uber</Bundle-Name>
-                        <Export-Package>com.algolia.search</Export-Package>
-                        <Import-Package>javax.net.ssl,javax.naming,*</Import-Package>
+                        <Bundle-Version>${project.version}</Bundle-Version>
+                        <Export-Package>com.algolia.search.*,com.algolia.search.org.apache.http.impl.nio.client</Export-Package>
```
Ant-hem: Maybe I am missing something but:

`com.algolia.search.org.apache.http.impl.nio.client` falls under `com.algolia.search.*` right?
Why do we need to specify `http.impl.nio...`?

algoliasearch-apache-uber/pom.xml
```diff
@@ -44,8 +44,13 @@
...
+                        <Import-Package>
+                            javax.net.ssl,javax.naming,!org.apache.avalon.framework.logger,!org.apache.log,*
```
Ant-hem, line 53: How did we find those import package?

Ant-hem (approved):
Thanks for your work! Huge enhancement for our AEM users.
My comments are curiosity questions and non-blocking.

algoliasearch-apache-uber/pom.xml
BenoitPerrot: @Ant-hem Haha! Precisely: because bundling (hence exploration/resolution/expansion of `com.algolia.search.*`) happens _before_ shading (hence before `org.apache...` becomes `com.algolia.search.org.apache...`). And (answering your second question) this is also why the full path is specified (`com.algolia.search.org.apache.http.impl.nio.client`).

The bundling plugin can't find `com.algolia.search.org.apache.http.impl.nio.client` at the time it runs: it does not exist at that time (the plugin is nice enough to add it to the `MANIFEST` `Export-Package` section nonetheless)

BenoitPerrot: @aseure here is what I was mentioning last time we talk, to explicit the relationship between this export and the shading below:

Add a property in a `<properties>` section above, eg:
```
<properties>
  <relocated.org.apache>com.algolia.search.org.apache</relocated.org.apache>
</properties>
```

Leverage that property in the configuration of the bundling package:
```
<Export-Package>com.algolia.search.*,${relocated.org.apache}.http.impl.nio.client</Export-Package>
```

Leverage that same property in the configuration of the shading plugin, in the `<relocation>` section below:
```
<pattern>org.apache</pattern>
<shadedPattern>${relocated.org.apache}</shadedPattern>
```

That way, it is made explicit by code that there is a relationship between the two. Someone reading the pom.xml later on will see that the two are linked (and that updating one require updating the other, etc). And as this is done by ~code (by opposition to a comment), it will support reorganization of the file. I think it is a small but desirable modification, considering the intricacy of the setup (bundling+shading+exposing a shaded package).

BenoitPerrot:
Actually I don't think this change is specific to AEM; it would be required on any OSGi engine. Let's be more clear about it in the commit title!

algoliasearch-apache-uber/pom.xml
aseure: @BenoitPerrot Thanks for all the answers you gave to @Ant-hem and for the recommendation to add the property. This is indeed a very good idea! Just adding it right now.

algoliasearch-apache-uber/pom.xml
aseure: By generating the uber JAR with `mvn package` and installing it to the OSGi/AEM instance and see which imports were missing there, one by one.

aseure:
> Actually I don't think this change is specific to AEM; it would be required on any OSGi engine. Let's be more clear about it in the commit title!

Good catch indeed @BenoitPerrot, thanks for noticing. I've amended the commit message as well.

algoliasearch-apache-uber/pom.xml
Ant-hem: Huge work! Thanks.

PR #711, opened 2020-07-15
feat(rule): implement multi-condition rules

### LINKED ISSUE #706: Multi-Condition Rules feature
This feature brings the support for multi-condition rules, which literally
means you can pass multiple conditions to trigger a single rule. The change
is backward-compatible and would ideally be shipped by mid-July or end-of-July
and it's gonna be deployed by June 19th for us to implement.

1. The feature is about exposing a new `conditions` array of rule conditions
in the rule object. The existing `condition` field remains for
backward-compatibility.
2. Also, please update the CTS like so: https://github.com/algolia/algoliasearch-client-specs-internal/pull/126/files.
3. Mark the `condition` key as deprecated.

aseure:
Close #706

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/rules/Rule.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/rules/Rule.java
@@ -14,6 +14,7 @@ public class Rule implements Serializable {
 
   private String objectID;
   private Condition condition;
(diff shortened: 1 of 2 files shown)


aallam (approved):
LGTM

Ant-hem (approved):
Sweet, thanks!

PR #712, opened 2020-07-15
test(api-keys): retry on restoreApiKeys upon '[404] Key already exists' error

### LINKED ISSUE #713: Implement retry until success for restore_api_key() call in CTS test
Because of the way Algolia is designed, performing a `restoreApiKey` may not
succeed because the previous `deleteApiKey` takes some time to get propagated
to all Algolia servers, giving us the following response: `[404] Key already
exists`.

To prevent that, we need to change the `testApiKeys` CTS test, if implemented,
to retry even on `[404] Key already exists` response.

aseure:
Close #713

### DIFF
--- a/algoliasearch-core/src/test/java/com/algolia/search/integration/client/ApiKeysTest.java
+++ b/algoliasearch-core/src/test/java/com/algolia/search/integration/client/ApiKeysTest.java
@@ -4,7 +4,9 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.algolia.search.SearchClient;
+import com.algolia.search.exceptions.AlgoliaApiException;
 import com.algolia.search.exceptions.AlgoliaRuntimeException;
+import com.algolia.search.integration.TestHelpers;
 import com.algolia.search.models.apikeys.*;
 import java.time.Duration;
 import java.time.Instant;
@@ -22,7 +24,7 @@ public abstract class ApiKeysTest {
   }
 
   @Test
-  void testApiKeys() {
+  void testApiKeys() throws Exception {
 
     ApiKey apiKeyToSend =
         new ApiKey()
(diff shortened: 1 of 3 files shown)


algoliasearch-core/src/test/java/com/algolia/search/integration/client/ApiKeysTest.java
```diff
@@ -62,7 +63,17 @@ void testApiKeys() {
     DeleteApiKeyResponse deleteApiKey = searchClient.deleteApiKeyAsync(apiKey).join();
     deleteApiKey.waitTask();
 
-    RestoreApiKeyResponse restoreApiKey = searchClient.restoreApiKeyAsync(apiKey).join();
+    RestoreApiKeyResponse restoreApiKey;
+    while (true) {
+      try {
+        restoreApiKey = searchClient.restoreApiKeyAsync(apiKey).join();
```
aallam: Not sure if its important in this case, but maybe a retry count logic here to avoid retrying indefinitely?

algoliasearch-core/src/test/java/com/algolia/search/integration/client/ApiKeysTest.java
```diff
@@ -62,7 +63,17 @@ void testApiKeys() {
...
+        break;
+      } catch (AlgoliaApiException e) {
+        if (e.getHttpErrorCode() == 404 && e.getMessage().equals("Key already exists")) {
```
Ant-hem: The engine doesn't commit on error messages :D But I guess that's fine for this test.

aseure: @Ant-hem I'm not sure of what you mean here?

algoliasearch-core/src/test/java/com/algolia/search/integration/client/ApiKeysTest.java
aseure: Since most of our `.waitTask()` methods are blocking without time limitation, I thought it wouldn't matter here for the test. But you're right, this is good practice anyway. I'll add a `retry()` test helper and use it here 👍

algoliasearch-core/src/test/java/com/algolia/search/integration/client/ApiKeysTest.java
nunomaduro: @aseure He meant that the API engine may change this message "Key already exists" in the future. But is actually fine,  because we are using it in tests here.

Ant-hem: Exactly!

aallam (approved):

PR #723, opened 2020-10-19
fix(ABTest): add missing fields in Variant

### LINKED ISSUE #721: Add missing fields in Variant.java
- Algolia Client Version: 3.X
- Language Version: All

### Description

The [Variant](https://github.com/algolia/algoliasearch-client-java-2/blob/master/algoliasearch-core/src/main/java/com/algolia/search/models/analytics/Variant.java) POJO seems to be missing lots of fields comparing to the other clients.

Let's update it with all the latest fields.

See C# struct as a reference:
https://github.com/algolia/algoliasearch-client-csharp/blob/master/src/Algolia.Search/Models/Analytics/Variant.cs

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #721 

## Describe your change

Add missing fields in `Variant` class.
Followed the [Kotlin implementation](https://github.com/algolia/algoliasearch-client-kotlin/blob/a5425e051fcef0a1dcbe65467ec884436008108b/src/commonMain/kotlin/com/algolia/search/model/response/ResponseVariant.kt#L22), seemed to be more complete.

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/analytics/Variant.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/analytics/Variant.java
@@ -28,17 +28,11 @@ public class Variant implements Serializable {
   // API.
   private String description;
 
-  // Average click position for the variant.
-  private Float averageClickPosition;
+  // Click through rate for the variant.
+  private Integer conversionCount;
 
   // Distinct click count for the variant.
   private Integer clickCount;
-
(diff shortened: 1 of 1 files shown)


Ant-hem (approved):
Thanks! On a side note, it looks like the API Key test is getting flaky..

PR #735, opened 2021-01-27
feat(dictionaries): dictionaries API implementation

aallam:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

Implement dictionaries API
- [x] `saveDictionaryEntries`
- [x] `replaceDictionaryEntries`
- [x] `deleteDictionaryEntries`
- [x] `clearDictionaryEntries`
- [x] `searchDictionaryEntries`
- [x] `setDictionarySettings`
- [x] `getDictionarySettings`
- [x] `waitAppTask`


algoliasearch-core/src/main/java/com/algolia/search/SearchClientBase.java
```diff
@@ -12,4 +12,8 @@ void waitTask(
       @Nonnull String indexName, long taskId, long timeToWait, RequestOptions requestOptions);
 
   void waitTask(@Nonnull String indexName, long taskId);
+
+  void waitAppTask(long taskId, long timeToWait, RequestOptions requestOptions);
```
aallam, line 16: Introducing `waitAppTask` to mark the difference between waiting for an indexing task vs application task.
The objective is to avoid misuse of this method for index task.

algoliasearch-apache/src/test/java/com/algolia/search/dictionary/DictionaryTest.java
```diff
@@ -0,0 +1,12 @@
+package com.algolia.search.dictionary;
+
+import com.algolia.search.IntegrationTestExtension;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith({IntegrationTestExtension.class})
+public class DictionaryTest extends com.algolia.search.integration.dictionary.DictionaryTest {
```
aimiebell, line 7: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names shouldn’t shadow simple name of superclass (NM_SAME_SIMPLE_NAME_AS_SUPERCLASS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-shouldn-t-shadow-simple-name-of-superclass-nm-same-simple-name-as-superclass).
This class has a simple name that is identical to that of its superclass, except that its superclass is in a different package (e.g., alpha.Foo extends beta.Foo). This can be exceptionally confusing, create lots of situations in which you have to look at import statements to resolve references and creates many opportunities to accidentally define methods that do not override methods in their superclasses.

tkrugg (approved):

PR #738, opened 2021-02-22
fix(osgi): add missing embedded classes to the exported packages

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Fix OSGi bundle:
* Add `com.algolia.search.com.fasterxml.jackson.annotation.JsonInclude.Include` to package exports.
* Exclude `javax.servlet` from the package imports.

### DIFF
--- a/algoliasearch-apache-uber/pom.xml
+++ b/algoliasearch-apache-uber/pom.xml
@@ -17,6 +17,7 @@
     </dependencies>
     <properties>
         <relocated.org.apache>com.algolia.search.org.apache</relocated.org.apache>
+        <relocated.com.fasterxml.jackson>com.algolia.search.com.fasterxml.jackson</relocated.com.fasterxml.jackson>
     </properties>
     <build>
         <plugins>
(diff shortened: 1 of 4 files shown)


algoliasearch-apache-uber/pom.xml
```diff
@@ -48,9 +49,14 @@
...
-                        <Export-Package>com.algolia.search.*,${relocated.org.apache}.http.impl.nio.client</Export-Package>
+                        <Export-Package>
+                            com.algolia.search.*,
+                            ${relocated.org.apache}.http.impl.nio.client,
+                            ${relocated.com.fasterxml.jackson}.annotation,
+                            ${relocated.com.fasterxml.jackson}.databind
+                        </Export-Package>
                         <Import-Package>
-                            javax.net.ssl,javax.naming,!org.apache.avalon.framework.logger,!org.apache.log,*
+                            javax.net.ssl,javax.naming,!org.apache.avalon.framework.logger,!org.apache.log,!javax.servlet,*
```
Ant-hem, line 59: question: why are we dropping `!javax.servlet`?

aallam: This first came up while doing my tests with Apache Karaf which does no provide `javax.servlet` (anymore?).
The bundle was not able to start because `javax.servlet` is a dependency and it is missing! 
However, the client does not actually need it to work! excluding it solved the issue.

Ant-hem (approved):
LGTM! Thanks for working on it.

PR #745, opened 2021-06-02
fix(rules): custom serialization for `TimeRange`

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Specify custom serialization for `TimeRange`, using Unix timestamps in seconds.

Rule's validity time ranges (from, until) only support integer Unix timestamps in seconds. 
The client's Jackson configuration serializes/deserializes those values using millisecond timestamps.
From Jackson's `JavaTimeModule`:

> For serialization, timestamps are written as fractional numbers (decimals), where the number is seconds and the decimal is fractional seconds, (...) If WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS is disabled, timestamps are written as a whole number of milliseconds.

### DIFF
--- a/algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
+++ b/algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
@@ -991,7 +991,7 @@ class JacksonParserTest {
     assertThat(json).isEqualTo("{\"from\":1619807400,\"until\":1630263239}");
 
     TimeRange retrieveTimeRange = Defaults.getObjectMapper().readValue(json, TimeRange.class);
-    assertThat(timerange.getFrom().isEqual(retrieveTimeRange.getFrom())).isTrue();
-    assertThat(timerange.getUntil().isEqual(retrieveTimeRange.getUntil())).isTrue();
+    assertThat(timerange.getFrom()).isEqualTo(retrieveTimeRange.getFrom());
+    assertThat(timerange.getUntil()).isEqualTo(retrieveTimeRange.getUntil());
   }
 }
--- a/algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
+++ b/algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
@@ -20,6 +20,7 @@ import com.algolia.search.models.rules.ConsequenceQuery;
 import com.algolia.search.models.rules.Edit;
 import com.algolia.search.models.rules.EditType;
 import com.algolia.search.models.rules.Rule;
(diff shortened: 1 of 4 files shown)


algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
```diff
@@ -974,4 +977,21 @@ void testConsequencePromotion() throws IOException {
...
+    TimeRange timerange =
+        new TimeRange(
+            OffsetDateTime.of(2021, 5, 1, 0, 0, 0, 0, offset),
+            OffsetDateTime.of(2021, 8, 30, 0, 23, 59, 0, offset));
+
+    String json = Defaults.getObjectMapper().writeValueAsString(timerange);
+    assertThat(json).isEqualTo("{\"from\":1619807400,\"until\":1630263239}");
+
+    TimeRange retrieveTimeRange = Defaults.getObjectMapper().readValue(json, TimeRange.class);
+    assertThat(timerange.getFrom().isEqual(retrieveTimeRange.getFrom())).isTrue();
```
Haroenv:
```suggestion
    assertThat(timerange.getFrom()).isEqual(retrieveTimeRange.getFrom());
```

should it be something more like this? it seems confusing to use isEqual & isTrue

algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
```diff
@@ -974,4 +977,21 @@ void testConsequencePromotion() throws IOException {
...
+    assertThat(timerange.getUntil().isEqual(retrieveTimeRange.getUntil())).isTrue();
```
Haroenv:
```suggestion
    assertThat(timerange.getUntil()).isEqual(retrieveTimeRange.getUntil());
```

Haroenv (approved):
I'm not sure if I understand this 100%, but from my POV this is correct

algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
aallam: Fixed! the old version of our testing library assertj wasn't behaving as expected.
Updating it allows the use of a better syntax :)

algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
aallam: Same as above

PR #747, opened 2021-06-23
feat(dynamic facets): facet ordering support

aallam:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

The new Facet Ordering feature providing the order of facet widgets and facet values.

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/SearchResult.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/SearchResult.java
@@ -1,5 +1,6 @@
 package com.algolia.search.models.indexing;
 
+import com.algolia.search.models.rules.RenderingContent;
 import com.algolia.search.util.AlgoliaUtils;
 import com.fasterxml.jackson.annotation.JsonSetter;
 import java.io.Serializable;
@@ -31,6 +32,7 @@ public class SearchResult<T> implements Serializable {
   private Boolean exhaustiveNbHits;
   private Integer appliedRelevancyStrictness;
   private Integer nbSortedHits;
+  private RenderingContent renderingContent;
 
   public Explain getExplain() {
     return explain;
@@ -128,6 +130,15 @@ public class SearchResult<T> implements Serializable {
 
   public SearchResult<T> setQueryAfterRemoval(String queryAfterRemoval) {
     this.queryAfterRemoval = queryAfterRemoval;
+    return this;
+  }
+
+  public RenderingContent getRenderingContent() {
+    return renderingContent;
+  }
+
+  public SearchResult<T> setRenderingContent(RenderingContent renderingContent) {
+    this.renderingContent = renderingContent;
     return this;
   }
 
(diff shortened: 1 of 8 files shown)


algoliasearch-core/src/main/java/com/algolia/search/models/indexing/SearchResult.java
```diff
@@ -131,6 +133,15 @@ public String getQueryAfterRemoval() {
     return this;
   }
 
+  public RenderingContent getRenderingContent() {
+    return renderingContent;
+  }
+
+  public SearchResult<T> setRenderingContent(RenderingContent renderingContent) {
```
Haroenv, line 140: I don't understand why this needs a public setter, but the other items seem to have it, so it's ok

algoliasearch-core/src/main/java/com/algolia/search/models/rules/FacetValuesOrder.java
```diff
@@ -0,0 +1,63 @@
...
+  public List<String> getOrder() {
+    return order;
+  }
+
+  public FacetValuesOrder setOrder(List<String> order) {
+    this.order = order;
+    return this;
+  }
+
+  public String getSortRemainingBy() {
```
Haroenv, line 30: can this be more restricted than string?

algoliasearch-core/src/main/java/com/algolia/search/models/rules/Consequence.java
```diff
@@ -60,9 +60,19 @@ public Consequence setFilterPromotes(Boolean filterPromotes) {
...
+    this.renderingContent = renderingContent;
+    return this;
+  }
+
   private ConsequenceParams params;
   private List<ConsequencePromote> promote;
   private Map<String, Object> userData;
   private List<Hide> hide;
   private Boolean filterPromotes;
+  private RenderingContent renderingContent;
```
Haroenv: renderingContent is actually _inside_ params in the consequence

algoliasearch-core/src/main/java/com/algolia/search/models/indexing/SearchResult.java
aallam: it is required by the serialization library (jackson)

algoliasearch-core/src/main/java/com/algolia/search/models/rules/FacetValuesOrder.java
aallam: We can, but this client's philosophy avoids introducing restrictions (this is observable in many other places in the codebase)

algoliasearch-core/src/main/java/com/algolia/search/models/rules/Consequence.java
aallam: Nice catch! fixed :)

Haroenv (approved):
looks consistent!

PR #749, opened 2021-06-28
fix(query): accessing empty nested `List<List<?>>`

### LINKED ISSUE #748: List could be empty
https://github.com/algolia/algoliasearch-client-java-2/blob/e36c48fc540e9218210a0858c4b39d04ebef315f/algoliasearch-core/src/main/java/com/algolia/search/util/QueryStringUtils.java#L72

If an empty list is sent then causes java.lang.IndexOutOfBoundsException in that line.

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | Fix #748 
| Need Doc update   | no

## Describe your change

* Add list emptiness check
* Remove unnecessary `null` check; `instanceof` returns `false` for nulls


Ant-hem (approved):
Nice catch!

PR #752, opened 2021-08-19
feat: recommend client implementation

### LINKED ISSUE #756: Recommend client
We need to add a new client to interact with the [Recommend](https://www.algolia.com/doc/rest-api/recommend/#get-recommendations) API. The API and implementation can be based on the [JavaScript client](https://github.com/algolia/algoliasearch-client-javascript/pull/1280).

aallam:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no
| Related Issue     | fix #756
| Need Doc update   | yes

## Describe your change

Recommend Client [API](https://www.algolia.com/doc/rest-api/recommend/#get-recommendations) implementation.

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/RecommendationsResult.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/RecommendationsResult.java
@@ -1,0 +1,15 @@
+package com.algolia.search.models.indexing;
+
+public class RecommendationsResult<T> extends SearchResult<T> {
+
+  private Integer score;
+
+  public Integer getScore() {
+    return score;
+  }
+
+  public RecommendationsResult<T> setScore(Integer score) {
+    this.score = score;
+    return this;
+  }
+}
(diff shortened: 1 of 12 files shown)


algoliasearch-core/src/main/java/com/algolia/search/models/recommend/RecommendationsOptions.java
```diff
@@ -0,0 +1,38 @@
...
+  Integer threshold;
+
+  /** The maximum number of recommendations to retrieve */
+  Integer maxRecommendations;
+
+  /** search parameters to filter the recommendations, */
+  Query queryParameters;
+
+  /** search parameters to use as fallback when there are no recommendations */
+  Query fallbackParameters;
```
Haroenv, line 37: technically not all parameters are allowed, but the api will fail so it's ok not to enforce that in types

Haroenv (approved):
looks great to me!

algoliasearch-core/src/main/java/com/algolia/search/models/indexing/RecommendationsResult.java
```diff
@@ -0,0 +1,15 @@
+package com.algolia.search.models.indexing;
+
+public class RecommendationsResult<T> extends SearchResult<T> {
+
+  private Integer score;
```
sbellone, line 5: Oh actually the `_score` is not at this level, but is present in all `hits`: https://www.algolia.com/doc/api-reference/api-methods/get-recommendations/?client=javascript#method-response-result

PR #753, opened 2021-08-20
fix(rules): automatic facet filter deserialization

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | yes/no

## Describe your change

Deserialize `AutomaticFacetFilterDeserializer` when it has only `facet` param.

### DIFF
--- a/algoliasearch-core/src/test/java/com/algolia/search/serialization/TestAutomaticFacetFilterDeserializer.java
+++ b/algoliasearch-core/src/test/java/com/algolia/search/serialization/TestAutomaticFacetFilterDeserializer.java
@@ -1,0 +1,27 @@
+package com.algolia.search.serialization;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import com.algolia.search.Defaults;
+import com.algolia.search.models.rules.AutomaticFacetFilter;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import org.junit.jupiter.api.Test;
+
+public class TestAutomaticFacetFilterDeserializer {
+
+  @Test
+  void testDeserializer() throws JsonProcessingException {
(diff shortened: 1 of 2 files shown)


algoliasearch-core/src/test/java/com/algolia/search/serialization/TestAutomaticFacetFilterDeserializer.java
```diff
@@ -0,0 +1,27 @@
...
+  @Test
+  void testDeserializer() throws JsonProcessingException {
+    String json = "{\"facet\":\"brand_name\"}";
+    AutomaticFacetFilter automaticFacetFilter =
+        Defaults.getObjectMapper().readValue(json, AutomaticFacetFilter.class);
+    assertThat(automaticFacetFilter.getFacet()).isEqualTo("brand_name");
+  }
+
+  @Test
+  void testDeserializerLegacy() throws JsonProcessingException {
```
Ant-hem, line 37: There is another possible case `["lastname","firstname"]` which is not covered by the tests. See:

https://github.com/algolia/algoliasearch-client-csharp/blob/f2c10db5f84a141f8e3ae54b9c30a8262db5441a/src/Algolia.Search.Test/Serializer/SerializerTest.cs#L183

Ant-hem (commented):
Looks good, let's add another test case and we are good to go :)

algoliasearch-core/src/test/java/com/algolia/search/serialization/TestAutomaticFacetFilterDeserializer.java
aallam: Thank you for the review! I've added the missing test 👍

Ant-hem (approved):

PR #758, opened 2021-10-14
fix(rules): deserialize optional filters

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | no

## Describe your change

Handle `optionalFilters` deserialization in the case of mixed formats (String and List)


algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
```diff
@@ -212,6 +212,16 @@ void testLegacyFiltersFormat(String input) throws IOException {
             Defaults.getObjectMapper().readValue(nestedArrayFilters, ConsequenceParams.class),
             input));
 
+    // Testing mixed case with array and string
+    // [["color:green","color:yellow"],"color:blue"]
+    String stringAndArrayFilters =
```
Ant-hem, line 217: We have the same issue with the C# client and Go, we probably not handle this use-case.

Ant-hem (approved):
Thanks for the fix 👍🏻

algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
aallam: True, I've checked the C# client, didn't see something related to this edge case

PR #761, opened 2021-10-26
fix(recommendation): `_score` field as member of `RecommendHit`

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | yes

## Describe your change

This fixes `_score` field, which is a property of `hits` not `results` as mentioned in the [doc](https://www.algolia.com/doc/rest-api/recommend/#get-recommendations).

_PS: this PR removes recommendation calls with `hits` as list of `Object`s since this can never be the case._


sbellone (approved):
Looks good! 🚀 
Don't forget the doc :)

PR #765, opened 2021-12-13
chore: dependencies check

aallam:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | no

## Describe your change

Adds dependencies check step to the build process using [Sonatype OSS Index](https://sonatype.github.io/ossindex-maven/enforcer-rules/).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -139,23 +139,25 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-enforcer-plugin</artifactId>
-                        <version>3.0.0</version>
+                        <version>3.0.0-M3</version>
+                        <dependencies>
+                            <dependency>
+                                <groupId>org.sonatype.ossindex.maven</groupId>
+                                <artifactId>ossindex-maven-enforcer-rules</artifactId>
+                                <version>3.1.0</version>
+                            </dependency>
+                        </dependencies>
                         <executions>
                             <execution>
-                                <id>cve-2021-44228</id>
+                                <id>vulnerability-checks</id>
                                 <phase>validate</phase>
                                 <goals>
                                     <goal>enforce</goal>
                                 </goals>
                                 <configuration>
                                     <rules>
-                                        <bannedDependencies>
-                                            <excludes>
-                                                <exclude>org.apache.logging.log4j:log4j-core:(,2.15.0)</exclude>
-                                            </excludes>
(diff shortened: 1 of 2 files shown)


sbellone (approved):

pom.xml
```diff
@@ -136,6 +136,32 @@
                             </execution>
                         </executions>
                     </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-enforcer-plugin</artifactId>
+                        <version>3.0.0-M3</version>
```
sbellone, line 142: I think you downgraded by mistake in your last changes, you originally had put the latest version:
```suggestion
                        <version>3.0.0</version>
```

aallam: `ossindex-maven-enforcer-rules` version `3.1.0` do not support `maven-enforcer-plugin` version `3.0.0` yet, had to downgrade to `3.0.0-M3` :/
Shouldn't be a big deal, since what matters is the rules themselves :)

sbellone: 😱 Oh no, they claim compatibility with 3+: https://sonatype.github.io/ossindex-maven/enforcer-rules/
What happens if you don't specify the version like in their example?

sbellone (approved):

pom.xml
aallam: Nothing really as far as we are concerned :D

sbellone: ^^ But like which version does it gets when you mvn install?

Aallam: We have two things here:
* ossindex (`3.1.0`): gets the [rules](https://ossindex.sonatype.org/) to be applied (_i.e vulnerable dependencies_).
* enforcer plugin (`3.0.0-M3`): checks if we have any dependency reported by the rules; the build fails if it is the case with an error message.

All this is applied to the build step only, and nothing is packaged to the final artifact :)

aallam: Oh, my bad! I didn't get your question at first glance! 
In their example, they must be using `dependencyManagement` for `maven-enforcer-plugin` too, but it is not mentioned in their example.
In the end, a version must be specified _(directly or indirectly)_.

sbellone: > In the end, a version must be specified (directly or indirectly).

Alright that's the part I was not sure about, thanks for confirming! 👍

PR #768, opened 2022-01-20
fix: copySettings func not calling expected base func

Ant-hem:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     

## Describe your change

The `copySettings` method was calling base method `copyIndex` instead of `copySettings`, leading to unexpected behavior: all scopes being copied instead of only the settings.


aallam (approved):
Nice catch! 🚀

Ant-hem:
It was reported by a user :)

PR #769, opened 2022-01-27
feat: TimeRange: deserialize time as long

BenoitPerrot:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | 
| Need Doc update   | no

## Describe your change

Homogenize JSON deserialization of TimeRange with serialization, using `long` for timestamps.

Some customers build rules with an end date very far away in the future, which cannot be captured as number of seconds from epoch with an int.


Ant-hem (approved):
Thanks!

PR #771, opened 2022-01-28
fix: optional filters deserialization

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | no

## Describe your change

Fixes `optionalFilters` deserialization for following special cases: 
* `["A:1", "B:2"]` ->   `[["A:1"], ["B:2"]]`
* `"A:1", "B:2"`  ->   `[["A:1"], ["B:2"]]`

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
@@ -29,12 +29,8 @@ class FiltersJsonDeserializer extends JsonDeserializer<List<List<String>>> {
 
     switch (currentToken) {
       case START_ARRAY:
-        List list = p.readValueAs(List.class);
-        if (list.stream().allMatch(String.class::isInstance)) { // are all elements strings?
-          result = Collections.singletonList(list);
-        } else {
-          result = buildFilters(list);
-        }
+        List<Object> list = p.readValueAs(List.class);
+        result = buildFilters(list);
         break;
       case VALUE_STRING:
         result = Collections.singletonList(Arrays.asList(p.getValueAsString().split(",")));
@@ -64,3 +60,4 @@ class FiltersJsonDeserializer extends JsonDeserializer<List<List<String>>> {
             .collect(Collectors.toList());
   }
 }
+
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
@@ -33,10 +33,8 @@ class FiltersJsonDeserializer extends JsonDeserializer<List<List<String>>> {
         result = buildFilters(list);
         break;
       case VALUE_STRING:
-        result =
-            Arrays.stream(p.getValueAsString().split(","))
-                .map(Collections::singletonList)
-                .collect(Collectors.toList());
+        String string = p.getValueAsString();
+        result = buildFilters(string);
         break;
       case VALUE_NULL:
         break;
@@ -48,6 +46,7 @@ class FiltersJsonDeserializer extends JsonDeserializer<List<List<String>>> {
     return result;
   }
 
+  /** Build filters from a list */
   @SuppressWarnings("unchecked")
   private List<List<String>> buildFilters(List list) {
     return (List<List<String>>)
@@ -61,5 +60,17 @@ class FiltersJsonDeserializer extends JsonDeserializer<List<List<String>>> {
                   }
                 })
             .collect(Collectors.toList());
+  }
+
+  /** Build filters from (legacy) string */
+  private List<List<String>> buildFilters(String string) {
+    if (string.startsWith("(") && string.endsWith(")")) {
+      String input = string.substring(1, string.length() - 1);
+      return Collections.singletonList(Arrays.asList(input.split(",")));
+    } else {
+      return Arrays.stream(string.split(","))
+          .map(Collections::singletonList)
+          .collect(Collectors.toList());
+    }
   }
 }
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
@@ -60,4 +60,3 @@ class FiltersJsonDeserializer extends JsonDeserializer<List<List<String>>> {
             .collect(Collectors.toList());
   }
 }
-
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
+
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
@@ -33,7 +34,10 @@ class FiltersJsonDeserializer extends JsonDeserializer<List<List<String>>> {
         result = buildFilters(list);
         break;
       case VALUE_STRING:
-        result = Collections.singletonList(Arrays.asList(p.getValueAsString().split(",")));
+        result =
+            Arrays.stream(p.getValueAsString().split(","))
+                .map(Collections::singletonList)
+                .collect(Collectors.toList());
         break;
       case VALUE_NULL:
         break;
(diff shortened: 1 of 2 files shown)


algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
```diff
@@ -182,7 +182,7 @@ void testLegacyFiltersFormat(String input) throws IOException {
     // Testing "one array" legacy filters => should be converted to "ORED" nested filters
     // [["color:green","color:yellow"]]
```
Ant-hem: IIUC that the expected result:

```suggestion
    // [["color:green"],["color:yellow"]]
```

algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
```diff
@@ -242,6 +242,14 @@ void assertOREDResult(List<List<String>> result) {
     assertThat(result.get(0)).containsSequence("color:yellow");
   }
 
+  void assertOREDListResult(List<List<String>> result) {
```
Ant-hem: IIUC the filters are `ANDs` instead of `ORs`. Can you confirm @godelized? :)  

```suggestion
  void assertANDEDListResult(List<List<String>> result) {
```

Ant-hem (commented):
thanks for the fix! 

I am not sure about the behavior of single string filter. 

👋🏻 @godelized do you remember how `"color:green,color:yellow"` should be translated? is it `[["color:green"],["color:yellow"]]`

https://github.com/algolia/algoliasearch-client-java-2/blob/20793bf208053e13692d43d4d9237eb480ffa9c1/algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java#L174-L176

corps-fini:
Thanks!

> I am not sure about the behavior of single string filter.
> 
> 👋🏻 @godelized do you remember how `"color:green,color:yellow"` should be translated? is it `[["color:green"],["color:yellow"]]`

It looks like it is the legacy format:
https://github.com/algolia/AlgoliaSaaS/blob/330cbefd4e3d36eba3565760a2e9ca96708c3ca4/algolia/api/json_parsers/DecodeJSONStringArray.cpp#L153-L181.
If I understand correctly the first level are separate groups (i.e. conjunctions) and inside-parenthesis items are assigned to the same group (i.e.) disjunctions.

So:
- `color:green,color:yellow` is equivalent to `["color:green", "color:yellow"]` and `[["color:green"], ["color:yellow"]]`.
- `(color:green, color:yellow)` is equivalent to `[["color:green", "color:yellow"]]`.

To be validated with a test :smile:.

corps-fini:
I did a pass on the tests :)

---

https://github.com/algolia/algoliasearch-client-java-2/pull/771/files#diff-b4728c7df2af885fb4a7070b3ce19d5902bf55c3fe38ef704f519a3ed8657beeR174-R180

- `color:green,color:yellow`
- for the engine it is equivalent to a and: `[["color:green"],["color:yellow"]]`
- the test should use `assertANDEDListResult` instead of `assertOREDResult`

---

https://github.com/algolia/algoliasearch-client-java-2/pull/771/files#diff-b4728c7df2af885fb4a7070b3ce19d5902bf55c3fe38ef704f519a3ed8657beeR182-R194

- `["color:green","color:yellow"]`
- for the engine it is equivalent to a and: `[["color:green"],["color:yellow"]]`
- the test uses `assertOREDListResult` but I didn't find the definition, but since it's a and, the test should use `assertANDEDListResult`

---

https://github.com/algolia/algoliasearch-client-java-2/pull/771/files#diff-b4728c7df2af885fb4a7070b3ce19d5902bf55c3fe38ef704f519a3ed8657beeR196-R204

Looks good :)

aallam:
> * `color:green,color:yellow`
> * for the engine it is equivalent to a and: `[["color:green"],["color:yellow"]]`

> * `["color:green","color:yellow"]`
> * for the engine it is equivalent to a and: `[["color:green"],["color:yellow"]]`

`color:green,color:yellow` and  `["color:green","color:yellow"]` are then equivalent! I've pushed a small changes to match the above.
Thank you @godelized for your help 🙌

algoliasearch-core/src/test/java/com/algolia/search/JacksonParserTest.java
```diff
@@ -171,18 +171,18 @@ void serializeFacetFilters() throws IOException {
   @ValueSource(strings = {"facetFilters", "optionalFilters", "tagFilters", "numericFilters"})
   void testLegacyFiltersFormat(String input) throws IOException {
```
Ant-hem, line 172: We also need a test for that legacy use-case:

> (color:green, color:yellow) is equivalent to [["color:green", "color:yellow"]]

aallam: True! I've added the support for this case too! :D

algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
```diff
@@ -63,4 +61,16 @@
                 })
             .collect(Collectors.toList());
   }
+
+  /** Build filters from (legacy) string */
+  private List<List<String>> buildFilters(String string) {
+    if (string.startsWith("(") && string.endsWith(")")) {
```
Ant-hem: Sorry for being nitpicky - I think we are missing on valid case, e.g.

```java
"(color:green,color:yellow),color:black" 
=> ["color:green,color:yellow","color:black"]
=> [["color:green","color:yellow"],["color:black"]]
```

^ can you confirm @godelized? :)

corps-fini: `(color:green,color:yellow),color:black` means `(color = green OR color = yellow) AND color = black`.

It is equivalent to the following:

- `(color:green,color:yellow),(color:black)`
- `[["color:green","color:yellow"],["color:black"]]`
- `[["color:green","color:yellow"],"color:black"]`

Ant-hem: Is `(color:green,color:yellow),color:black` also a valid syntax? Are the parenthesis mandatory for the second group?

aallam: Is it also possible to have nested parentheses, like `(color:green,(color:yellow,color:blue)),color:pink`? if yes, should we support it?

corps-fini: > Is (color:green,color:yellow),color:black also a valid syntax?

Yes this is valid :)
See the legacy parsing function: https://github.com/algolia/AlgoliaSaaS/blob/330cbefd4e3d36eba3565760a2e9ca96708c3ca4/algolia/api/json_parsers/DecodeJSONStringArray.cpp#L153-L181.
The parenthesis are completely optional an denotes a disjunction.

> Is it also possible to have nested parentheses, like (color:green,(color:yellow,color:blue)),color:pink?

It is, I checked the code and tested it on my app but it does not work: when parsing the nested inner `)` the code does reset the state `insideGroup` :hot_face:.
I think we should reject inner parenthesis in the Java client and fix the engine.

aallam: > I think we should reject inner parenthesis in the Java client and fix the engine.

Since the legacy form can only come from the engine, and users can't provide such syntax, do we need the client to reject it?

Ant-hem: IMO it's the responsibility of the API to validate the payload, we shouldn't handle that case here. As @Aallam said, we'll retrieve rules already saved, meaning that the filters were validated by the API when saving the rule beforehand.

Ant-hem (approved):
thanks for the patch @Aallam, we now cover all the use-cases. :)

algoliasearch-core/src/main/java/com/algolia/search/models/indexing/FiltersJsonDeserializer.java
corps-fini: Makes sense but how do you deal with `(((A:1),B:2),(C:3),D:4)`?

Ant-hem: Maybe I am missing something - but in this case we have inner parenthesis `(A:1)` and hence this filter shouldn't be validated by the API, right?

corps-fini: If I understand correctly we convert the filter string to the JSON array format: https://github.com/algolia/algoliasearch-client-java-2/pull/771/files#diff-9135e60921839756742dc454f4b063cefc2f1e574560ca6ea51adb77ca28460bR35-R38.
Did I miss something?
So I am afraid that we'd convert `(((A:1),B:2),(C:3),D:4)` to something wrong.

corps-fini (approved):
Let's go :rocket:.
The only concern I have is that if a client sends a weird filter string like `(((A:1),B:2),(C:3),D:4)`, the engine is going to accept it (but the behavior is not correct imo). However when fetching it with the Java client, it will convert it to something different.
But I guess it's an edge case of something that is already a bug so its fine :+1:.

PR #772, opened 2022-02-03
feat: add default utility classes for `java-net`

### LINKED ISSUE #763: Readme example is only applicable when using apache http layer
- Algolia Client Version: 3.16.1
- Language Version: J11

### Description
The readme states apache or the java native client can be used, then follows with an example that only works with apache. DefaultSearchClient doesn't exist in algoliasearch-java-net.
The README could be adjusted, but ideally the default implementation of the SearchClient should be available and mostly identical in both variants.

### Steps To Reproduce
Try to follow the readme with algoliasearch-java-net.

aallam:
| Q                 | A
| Bug fix?          | no
| New feature?      | no 
| BC breaks?        | no     
| Related Issue     | Fixes #763
| Need Doc update   | no

## Describe your change

Have the same utility classes in both `apache` and `java-net` modules for clients instantiation:
* `DefaultSearchClient`
* `DefaultRecommendClient`
* `DefaultPersonalizationClient`
* `DefaultInsightsClient`
* `DefaultAnalyticsClient`

_Pitfall: using both `apache` and `java-net` at the same time may cause classes duplication errors._

### DIFF
--- a/algoliasearch-java-net/src/main/java/com/algolia/search/DefaultAnalyticsClient.java
+++ b/algoliasearch-java-net/src/main/java/com/algolia/search/DefaultAnalyticsClient.java
@@ -1,0 +1,59 @@
+package com.algolia.search;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Algolia's REST analytics client that wraps an instance of the transporter. which wraps the HTTP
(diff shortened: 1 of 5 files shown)


Ant-hem (approved):
Look for the good 👍🏻

PR #774, opened 2022-03-11
fix(HttpRequester): use body supplier

### LINKED ISSUE #773: Retry strategy: `HttpRequest`'s empty body
Users might get 400 errors indicating that there was no content in the body despite sending a request with body content. 

The same `HttpRequest` object, which stores the request body as an `InputStream`, is used for all request attempts.

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | Fix #773   
| Need Doc update   | no

## Describe your change

Instead of using the same `InputStream`, we use a `Supplier<InputStream>`, which produces a new `InputStream` at every retry attempt.


Ant-hem (approved):
Thanks nice catch. We had the same issue with the C# client actually 😓

PR #775, opened 2022-03-17
fix: add cipher suites for TLSv1.3 and HTTP/2

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | no

## Describe your change

In Java 11 (at least until `11.0.14`), when the API is configured with TLSv1.3 and HTTP/2, connections using them can not be established, this happens because no available cipher suite for TLSv1.3 can be found by default.

_PS: this issue does not appear in newer Java versions (i.g. Java 17) or by downgrading to HTTP/1.1_

### DIFF
--- a/algoliasearch-java-net/src/main/java/com/algolia/search/JavaNetHttpRequester.java
+++ b/algoliasearch-java-net/src/main/java/com/algolia/search/JavaNetHttpRequester.java
@@ -4,9 +4,6 @@ import com.algolia.search.exceptions.AlgoliaRuntimeException;
 import com.algolia.search.models.HttpRequest;
 import com.algolia.search.models.HttpResponse;
 import com.algolia.search.util.HttpStatusCodeUtils;
-
-import javax.annotation.Nonnull;
-import javax.net.ssl.SSLParameters;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.ProxySelector;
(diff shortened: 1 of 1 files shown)


algoliasearch-java-net/src/main/java/com/algolia/search/JavaNetHttpRequester.java
```diff
@@ -40,12 +46,17 @@ public JavaNetHttpRequester(@Nonnull ConfigBase config) {
         HttpClient.newBuilder()
             .executor(config.getExecutor())
             .version(HttpClient.Version.HTTP_2)
+            .sslParameters(sslParametersCiphers())
```
Ant-hem, line 49: Does `sslParameters` append to the default value of the JVM or does it completely override them?

aallam: After checking it is overriding them; and I've discovered another workaround: setting `SSLParameters` protocols to `null`! 

This made me think of another approach to do this: 
since this issue is related to a specific Java version (_Java 11 up until 11.0.14, and might be fixed later_), we can expose a customizable `HttpClient.Builder` (like for apache http client), this will allow the users to implement workarounds for this issue and any future ones, introduce the workaround for the  Java 11 tests, and run test against latest version (Java 17 for example).

wdyt ?

Ant-hem: Sounds like a great idea to me :)

aallam: Nice! I will close this PR then, I've opened another one instead.

PR #776, opened 2022-03-17
fix: java http client requester with `HttpClient.Builder`

aallam:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | n/A
| Need Doc update   | no

## Describe your change

Adds new constructor to `JavaNetHttpRequester(ConfigBase, HttpClient.Builder)`, giving control over the http client settings.


algoliasearch-java-net/src/main/java/com/algolia/search/JavaNetHttpRequester.java
```diff
@@ -36,12 +36,25 @@
...
+            : HttpClient.newBuilder());
+  }
+
+  /**
+   * Build the reusable instance of httpClient with the given configuration.
+   *
+   * @param config HTTPClient agnostic Algolia's configuration
+   * @param builder Builder for {@linkplain HttpClient HTTP Clients}
+   */
+  public JavaNetHttpRequester(@Nonnull ConfigBase config, HttpClient.Builder builder) {
```
Ant-hem, line 52: IIUC customers encountering the SSL issues will have to instantiate the Java client with a 'custom' builder that's correct as the default values won't be working for them?

Ant-hem (approved):
Nice addition to the codebase 👍🏻 thanks.

algoliasearch-java-net/src/main/java/com/algolia/search/JavaNetHttpRequester.java
aallam: Yes exactly, since the workaround depend on their Java version and environment.

PR #785, opened 2023-05-25
fix(core): http request timeout overflow

### LINKED ISSUE #784: Invalid duration exception on calculating request timeout
- Algolia Client Version: 3.16.6
- Language Version: Java

### Description
Hey,
We implemented a service, that polls messages from a Kafka topic, and  after a data transformation it writes them to our Algolia indices. 
It uses BatchOperations, so for writing data our implementation calls the multipleBatchAsync method on the SearchClient. 
The integration works mostly well, but sometimes we are experiencing a really strange issue. 
We use to have this kind of exceptions in our monitoring system: _"IllegalArgumentException (Invalid duration: PT-50H-39M-27.552S)"_, what is pretty strange, because the SDK tries to use a negative value as request timeout. (Please find the stack trace attached.)

After a short investigation, our best assumption is that the issue occurs, when the Java SDK (HttpTransport::executeWithRetry) retries the request, because of a timeout. 
In this case it increases the timeout on the request, that is stored in an int field on the corresponding  HttpRequest object.
During the further steps  the logic tries to convert this int field to a Duration object, and that is the point, where the issue comes up, as you can see in the attached stack trace.

So our assumption is, that the int value can overflow if we have too many retries.

Our assumption may not be correct, but could you please check it, and help us solving our issue?
Thanks.
[algolia_stacktrace_230504.txt](https://github.com/algolia/algoliasearch-client-java-2/files/11395379/algolia_stacktrace_230504.txt)

aallam:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | Fix #784
| Need Doc update   | no

## Describe your change

This PR modifies `HttpRequest#incrementTimeout` method to prevent integer overflow. In case of overflow, timeout is set to `Integer.MAX_VALUE`.

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/models/HttpRequest.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/models/HttpRequest.java
@@ -111,7 +111,12 @@ public class HttpRequest {
   }
 
   public void incrementTimeout(int retryCount) {
-    this.timeout *= (retryCount + 1);
+    // check for overflow
+    if (timeout > 0 && retryCount > Integer.MAX_VALUE / timeout - 1) {
+      timeout = Integer.MAX_VALUE;
+    } else {
+      timeout *= (retryCount + 1);
+    }
   }
 
   public Supplier<InputStream> getBodySupplier() {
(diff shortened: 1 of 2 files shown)


Fluf22 (approved):
Nice 👍🏻

PR #799, opened 2026-06-04
fix(java): ensure waitTask exception propagation

eric-zaharia:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no
| Related Issue     | [API-473]
| Need Doc update   | no

Jira ticket: [API-473]

## Describe your change

`TaskUtils.waitTask` polls task status in a `while (true)` loop and, on `InterruptedException` / `ExecutionException` from `getTask`, silently `break`ed out of the loop. Because `waitTask` returns `void`, a swallowed exception was indistinguishable from a genuine `"published"` confirmation.

This PR replaces the silent `break` with proper exception propagation, mirroring the `LaunderThrowable` pattern already used throughout the client:

- `InterruptedException` → restore the thread's interrupt flag, then rethrow as `AlgoliaRuntimeException`.
- `ExecutionException` → unwrap the real cause via `LaunderThrowable.launder`, surfacing the underlying `AlgoliaApiException` / `AlgoliaRetryException` / `AlgoliaRuntimeException`.

After the change the loop has only two exits: a real `"published"` status (`return`) or a thrown exception. It can no longer return as if a task completed when it did not.

If `waitTask` silently exited after the **copy** step (a transient network error, or `getTask` being retried on a host that had not yet registered the task), `replaceAllObjects` proceeded as though the copy had completed:

1. settings / rules / synonyms were **not** confirmed copied into the tmp index;
2. records were batched into the tmp index;
3. the tmp index was atomically moved over the main index.

**Result:** the main index keeps its records but loses its settings, rules, and synonyms (reset to defaults) — a silent partial wipeout, with no error surfaced to the caller. This is exactly the incident reported in [CR-11430].

This bug is unique to the Java **v3** (legacy) client — v4 and every other legacy client propagate the exception. v4 is already safe via `TaskUtils.retryUntil` (no `try/catch` around the supplier) plus a `replaceAllObjects` cleanup wrapper.

[API-473]: https://algolia.atlassian.net/browse/API-473
[CR-11430]: https://algolia.atlassian.net/browse/CR-11430

### DIFF
--- a/algoliasearch-core/src/main/java/com/algolia/search/TaskUtils.java
+++ b/algoliasearch-core/src/main/java/com/algolia/search/TaskUtils.java
@@ -3,6 +3,7 @@ package com.algolia.search;
 import com.algolia.search.exceptions.AlgoliaApiException;
 import com.algolia.search.exceptions.AlgoliaRetryException;
 import com.algolia.search.exceptions.AlgoliaRuntimeException;
+import com.algolia.search.exceptions.LaunderThrowable;
 import com.algolia.search.models.RequestOptions;
 import com.algolia.search.models.common.TaskStatusResponse;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


Fluf22 (approved):
Thank you

